PR #364, opened 2025-06-14
Add enhanced sensitive data handling

kuzaxak:
Functionality adds similar to [helm-diff][1] style of diffing sensitive
data.

Previously, NELM handled sensitive data with an all-or-nothing approach:
- Resources marked as sensitive (via `werf.io/sensitive: "true"`) were
  completely hidden
- Secrets were entirely hidden, providing no visibility into structural changes
- No way to selectively redact only sensitive fields while showing the rest

This made it difficult to:
- Track changes in Secret keys or metadata
- See the size of sensitive data changes
- Selectively protect only truly sensitive fields in complex resources

Added support for `werf.io/sensitive-paths` annotation that accepts a
JSONPath expressions:

```yaml
metadata:
 annotations:
  werf.io/sensitive-paths: $.spec.template.spec.containers[*].env[?(@.name=='API_KEY')].value,$.data.password
```

Sensitive values are replaced with size information:
```
password: SENSITIVE (len 12 bytes)
```

I used [ojg][2] library as JSONPath as kube client do not support
native mutation for complex structures like slices.

[1]: https://github.com/databus23/helm-diff
[2]: https://github.com/ohler55/ojg

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ We consider Nelm production-ready, since 95% of the Nelm codebase basically is t
     - [Annotation `<id>.external-dependency.werf.io/resource`](#annotation-idexternal-dependencywerfioresource)
     - [Annotation `<id>.external-dependency.werf.io/name`](#annotation-idexternal-dependencywerfioname)
     - [Annotation `werf.io/sensitive`](#annotation-werfiosensitive)
+    - [Annotation `werf.io/sensitive-paths`](#annotation-werfiosensitive-paths)
     - [Annotation `werf.io/track-termination-mode`](#annotation-werfiotrack-termination-mode)
     - [Annotation `werf.io/fail-mode`](#annotation-werfiofail-mode)
     - [Annotation `werf.io/failures-allowed-per-replica`](#annotation-werfiofailures-allowed-per-replica)
@@ -427,6 +428,21 @@ Example: `werf.io/sensitive: "true"`
 
 Don't show diffs for the resource.
 
+The behavior of this annotation depends on the `NELM_FEAT_FIELD_SENSITIVE` feature gate:
+- **Without feature gate (default):** Hides the entire resource content
+- **With feature gate:** Redacts only common sensitive fields (`data.*`, `stringData.*`) instead of hiding the entire resource
+
+#### Annotation `werf.io/sensitive-paths`
+
+Format: `JSONPath1,JSONPath2,...` \
+Example: `werf.io/sensitive-paths: "$.spec.template.spec.containers[*].env[*].value,$.data.*"`
+
+Allows fine-grained control over which specific fields should be redacted in diffs using JSONPath expressions. Multiple paths can be specified as a comma-separated list.
+
+This provides precise control over sensitive data redaction, allowing you to hide only specific sensitive fields (like passwords, API keys, etc.) rather than the entire resource, making diffs more useful while still protecting sensitive information.
+
+*Annotation precedence:* `werf.io/sensitive-paths` has highest priority, over  `werf.io/sensitive: "true"`
+
 #### Annotation `werf.io/track-termination-mode`
 
 Format: `WaitUntilResourceReady|NonBlocking` \
@@ -599,6 +615,21 @@ nelm release install -n myproject -r myproject
 ```
 
 Every few seconds print stack traces of all goroutines. Useful for debugging purposes.
+
+#### Env variable `NELM_FEAT_FIELD_SENSITIVE`
+
+Example:
+```shell
+export NELM_FEAT_FIELD_SENSITIVE=true
+nelm release plan install -n myproject -r myproject
+```
+
+Changes the behavior of the `werf.io/sensitive` annotation and default Secret handling:
+
+- **Without feature gate (default):** `werf.io/sensitive: "true"` and Secrets without annotations hide the entire resource content
+- **With feature gate:** `werf.io/sensitive: "true"` and Secrets without annotations hide only `data.*` and `stringData.*` fields
+
+Note: The `werf.io/sensitive-paths` annotation works regardless of this feature gate setting.
 
 ### More information
 
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 	github.com/jellydator/ttlcache/v3 v3.1.1
 	github.com/looplab/fsm v1.0.2
 	github.com/moby/term v0.5.0
+	github.com/ohler55/ojg v1.26.7
 	github.com/onsi/ginkgo/v2 v2.20.1
 	github.com/onsi/gomega v1.36.0
 	github.com/pkg/errors v0.9.1
@@ -32,6 +33,7 @@ require (
 	github.com/sourcegraph/conc v0.3.0
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.10.0
 	github.com/tidwall/sjson v1.2.5
 	github.com/wI2L/jsondiff v0.5.0
 	github.com/werf/3p-helm v0.0.0-20250609150428-130783e0dc18
@@ -135,6 +137,7 @@ require (
 	github.com/opencontainers/image-spec v1.1.0 // indirect
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/pjbgf/sha1cd v0.3.0 // indirect
+	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/prometheus/client_golang v1.19.0 // indirect
 	github.com/prometheus/client_model v0.6.0 // indirect
 	github.com/prometheus/common v0.48.0 // indirect
--- a/internal/resource/sensitive.go
+++ b/internal/resource/sensitive.go
@@ -1,0 +1,174 @@
+package resource
+
+import (
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/ohler55/ojg/jp"
+	"github.com/samber/lo"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/werf/nelm/pkg/featgate"
+)
+
+const (
+	HideAll = "$$HIDE_ALL$$"
+)
+
+type SensitiveInfo struct {
+	IsSensitive    bool
+	SensitivePaths []string
+}
+
+func IsSensitive(groupKind schema.GroupKind, annotations map[string]string) bool {
+	info := GetSensitiveInfo(groupKind, annotations)
+	return info.IsSensitive
+}
+
+func GetSensitiveInfo(groupKind schema.GroupKind, annotations map[string]string) SensitiveInfo {
+	// Check for werf.io/sensitive-paths (comma-separated)
+	if _, value, found := FindAnnotationOrLabelByKeyPattern(annotations, annotationKeyPatternSensitivePaths); found {
+		paths := ParseSensitivePaths(value)
+		if len(paths) > 0 {
+			return SensitiveInfo{IsSensitive: true, SensitivePaths: paths}
+		}
+	}
+
+	useNewBehavior := featgate.FeatGateFieldSensitive.Enabled() || featgate.FeatGatePreviewV2.Enabled()
+
+	// Check for werf.io/sensitive annotation
+	if _, value, found := FindAnnotationOrLabelByKeyPattern(annotations, annotationKeyPatternSensitive); found {
+		sensitive := lo.Must(strconv.ParseBool(value))
+		if sensitive {
+			if useNewBehavior {
+				// V2 behavior: only hide data.* and stringData.*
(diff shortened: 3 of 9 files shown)


kuzaxak:
@ilya-lesikov Hey, I hope it's alright to propose a change without causing any issues. We endeavoured to use Nelm, but the existing sensite diff functionality is rather limited.

ilya-lesikov:
@kuzaxak Hi, thanks for the PR. I'll look into it. Would've been better to create an issue first so we decide on the design, but whatever.

kuzaxak:
> @kuzaxak Hi, thanks for the PR, looks interesting. I'll look into it. Would've been better to create an issue first so we decide on the design, but whatever.

Thanks! I was thinking that it might be better to reuse existing annotation and determine, based on the type, if it's a bool, then use the hide all function; if it's a string that can be parsed into an array, then treat it as a JSON path.

ilya-lesikov:
@kuzaxak

Hi, skimmed through it, love the feature, thank you. Can we make a few design changes?

I think this feature is so flexible that it makes "werf.io/sensitive" kinda obsolete. How about we do this:
1. Add new featgate NELM_FEAT_FIELD_SENSITIVE=true via [featgate package](https://github.com/x-qdo/nelm/blob/ece6f8cd84f43baeb601675e7a2e2711036817b9/pkg/featgate/feat.go#L30-L33).
1. NELM_FEAT_FIELD_SENSITIVE=true will be the default in v2, so let's activate NELM_FEAT_FIELD_SENSITIVE by enabling [NELM_FEAT_PREVIEW_V2](https://github.com/x-qdo/nelm/blob/ece6f8cd84f43baeb601675e7a2e2711036817b9/pkg/featgate/feat.go#L35-L38).
1. Can we change the format a bit? `werf.io/sensitive-paths: "data.password,spec.template.spec.containers.*.env.*.value"`, in line with all the other annotations. `\,` should escape the comma, so we don't split entries on it.
1. If NELM_FEAT_FIELD_SENSITIVE=true and the resource has `werf.io/sensitive: "true"` annotation, then it works as `werf.io/sensitive-paths: "spec.*,data.*"`, otherwise current behavior
1. If NELM_FEAT_FIELD_SENSITIVE=true and the resource kind is Secret and it has no `werf.io/sensitive` or `werf.io/sensitive-paths` annotations, then it works as `werf.io/sensitive-paths: "data.*"`, otherwise current behavior
1. Let's use the actual JSONPath parser, for example "k8s.io/client-go/util/jsonpath", which we already use in werf/kubedog. Grok returned me this PoC, no idea if this works, but the general idea is right: https://pastebin.com/1bSFqza2 If "k8s.io/client-go/util/jsonpath" doesn't fit we can find something else.
1. Sometimes we replace "REDACTED" with "REDACTED" or "REDACTED (len 5 bytes)" with "REDACTED (len 5 bytes)", both of which produce no diff, which results in `<hidden insignificant changes>`. How about we compute SHA256 hash and truncate it to 12 chars, then show this: `SENSITIVE (5 bytes, e0c9035898dd)`. We also need to do this for slices and maps: do len(slice/map), then marshal the slice/map into the JSON string and get its SHA256, then show this: `SENSITIVE (10 entries, e0c9035898dd)`.
1. Also make sure that when type of a value changes (e. g. string -> slice) or when value is deleted/added as a whole, then the correct diffs are produced.

That's quite a few changes, but if this is done I'll merge it. I'll probably need to add CONTRIBUTORS.md and mention that it's best if we decide on the design first, since it's not easy to come up with it from the get-go. Pretty close overall, thanks again.

kuzaxak:
@ilya-lesikov hey, could you please check it?

README.md
```diff
@@ -600,6 +620,18 @@ nelm release install -n myproject -r myproject
 
 Every few seconds print stack traces of all goroutines. Useful for debugging purposes.
 
+#### Env variable `NELM_FEAT_FIELD_SENSITIVE`
+
+Example:
+```shell
+export NELM_FEAT_FIELD_SENSITIVE=true
+nelm release install -n myproject -r myproject
```
ilya-lesikov:
```suggestion
nelm release plan install -n myproject -r myproject
```

README.md
```diff
@@ -427,6 +428,25 @@ Example: `werf.io/sensitive: "true"`
 
 Don't show diffs for the resource.
 
+With the `NELM_FEAT_FIELD_SENSITIVE` feature gate enabled, when set to `true`, this annotation will redact common sensitive fields (`spec.*`, `data.*`, `stringData.*`) instead of hiding the entire resource.
+
+#### Annotation `werf.io/sensitive-paths`
+
+Format: `JSONPath1,JSONPath2,...` \
+Example: `werf.io/sensitive-paths: "$.spec.template.spec.containers[*].env[?(@.name=='API_KEY')].value,$.data.password"`
```
ilya-lesikov:
```suggestion
Example: `werf.io/sensitive-paths: "$.spec.template.spec.containers[*].env[*].value,$.data.*"`
```

README.md
```diff
@@ -427,6 +428,25 @@ Example: `werf.io/sensitive: "true"`
...
+
+This provides precise control over sensitive data redaction, allowing you to hide only specific sensitive fields (like passwords, API keys, etc.) rather than the entire resource, making diffs more useful while still protecting sensitive information.
+
+**Annotation precedence:**
+
+1. `werf.io/sensitive-paths` - highest priority, if present and non-empty, these JSONPath expressions are used
+2. `werf.io/sensitive: "true"` - medium priority, redacts common fields (`data.*`, `stringData.*`)  
+3. Default behavior for `v1/Secret` resources - lowest priority, redacts `data.*` and `stringData.*` fields
+
+If `werf.io/sensitive: "false"` is explicitly set, no sensitive redaction is applied regardless of resource type.
```
ilya-lesikov: They'll figure it out themselves, if really needed. Ultimately, this will be what the user does: if plan shows something sensitive, the user goes and adds `werf.io/sensitive-paths`, that's it.
```suggestion
```

internal/resource/sensitive.go
```diff
@@ -0,0 +1,188 @@
...
+	SensitivePaths []string
+}
+
+func IsSensitive(groupKind schema.GroupKind, annotations map[string]string) bool {
+	info := GetSensitiveInfo(groupKind, annotations)
+	return info.IsSensitive
+}
+
+func GetSensitiveInfo(groupKind schema.GroupKind, annotations map[string]string) SensitiveInfo {
+	if featgate.FeatGateFieldSensitive.Enabled() || featgate.FeatGatePreviewV2.Enabled() {
```
ilya-lesikov: I think `werf.io/sensitive-paths` can be allowed by default, without FeatGate. I was only worried about breaking current behavior of `werf.io/sensitive` and Secret defaults. Allowing `werf.io/sensitive-paths` by default doesn't break anything.

internal/resource/sensitive.go
```diff
@@ -0,0 +1,188 @@
...
+					"namespace": unstruct.GetNamespace(),
+				},
+			}}
+		}
+
+		x, err := jp.ParseString(pathExpr)
+		if err != nil {
+			// path already was validated
+			continue
+		}
```
ilya-lesikov:
```suggestion
		x := lo.Must(jp.ParseString(pathExpr))
		
```

go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	github.com/jellydator/ttlcache/v3 v3.1.1
 	github.com/looplab/fsm v1.0.2
 	github.com/moby/term v0.5.0
+	github.com/ohler55/ojg v1.26.7
```
ilya-lesikov, line 27: Can we use this package instead? https://github.com/werf/kubedog/blob/654a0802b36e6df9630f6ac5653314f895a90550/pkg/utils/json.go#L8

ilya-lesikov: Alright, I see

internal/resource/sensitive.go
kuzaxak: Okay, now it makes sense. I misunderstood you in the previous message.

So, in short: the feature flag only affects current annotation behaviour and behaviour without annotation for secrets. Is the new functionality available as is, without a feature flag?

README.md
kuzaxak: I didn't get you. Annotations can overlap each other, and here we are documenting which one has the highest priority. 

As an example, what happens if the user sets `werf.io/sensitive: "false"` and `werf.io/sensitive-paths` at the same time?

README.md
kuzaxak: I think as we are supporting full jsonpath spec it makes sense to give a complex example in a doc, easy example can be added in the end, like `werf.io/sensitive-paths: "$.spec.template.spec.containers[*].env[?(@.name=='API_KEY')].value,$.data.*"

However, the primary purpose of this annotation is to mark specific fields in the manifest as sensitive, using the annotation' werf.io/sensitive' for simple `.data.*` fields. 

Or do you want to have only one annotation in the end?

ilya-lesikov: I wanted this annotation to effectively supersede the werf.io/sensitive annotation. It's just superior.

And in regards to usefulness of `.*`, then frankly I think for most people it would be more practical to just `.*` fields that can be compromising. Otherwise you need to know and remember to add to `werf.io/sensitive-paths` any new secret that is added to, let's say, container env vars. Knowing that sometimes there are hundreds of env vars, many of them are sensitive, and there are might be multiple people editing these env vars, it's hard to expect some of these env vars not to be leaked here and there. So I'd pick better security over a little bit better diffs for our docs.

README.md
ilya-lesikov: Well, I guess you can briefly mention that werf.io/sensitive-paths has higher priority than werf.io/sensitive. Ultimately I'd like to see werf.io/sensitive annotation go to some Deprecated/Legacy section of the docs. werf.io/sensitive was just a quick fix for a complicated problem.

internal/resource/sensitive.go
ilya-lesikov: Yes, it should only change `werf.io/sensitive` behavior and the default Secret behavior (which is an equivalent to `werf.io/sensitive: true`). The `werf.io/sensitive-paths` annotation though should be available to everyone, with or without feature gate enabled.

kuzaxak: Gotcha! I will fix the behaviour tomorrow. Thanks for the explanations!

internal/resource/sensitive.go
kuzaxak: Fixed

internal/resource/sensitive.go
kuzaxak: Implemented

README.md
kuzaxak: Replaced with a simple one-sentence doc.

README.md
kuzaxak: Updated with your proposal

ilya-lesikov (changes requested):
Tried it out, the only issue I found is this:
```shell
❯ cat chart/templates/test.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: to-create
  annotations:
    werf.io/sensitive-paths: '$.data[*]'
data:
  key1: {{ uuidv4 }}
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: to-update
  annotations:
    werf.io/sensitive-paths: '$.data[*]'
data:
  key1: {{ uuidv4 }}
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: to-delete
  annotations:
    werf.io/sensitive-paths: '$.data[*]'
data:
  key1: {{ uuidv4 }}

❯ nelm release install -n jkldfjkdkjjkdk -r jkldfj .helm

❯ NELM_FEAT_FIELD_SENSITIVE=false nelm release plan install -n jkldfjkdkjjkdk -r jkldfj .helm
Planning release install "jkldfj" (namespace: "jkldfjkdkjjkdk")

┌ Create ConfigMap/to-create
│ + apiVersion: v1
│ + kind: ConfigMap
│ + metadata:
│ +   name: to-create
│ +   namespace: ""
└ Create ConfigMap/to-create

┌ Update ConfigMap/to-delete
│ <hidden insignificant changes>
└ Update ConfigMap/to-delete

┌ Update ConfigMap/to-update
│ <hidden insignificant changes>
└ Update ConfigMap/to-update

Planned changes summary for release "jkldfj" (namespace: "jkldfjkdkjjkdk"):
- create: 1 resource(s)
- update: 2 resource(s)
```
Otherwise works great. If you push the fix, please don't force-push it, just add new commits, this way it's easier for me to review only the new changes. I'll squash all the extra commits just before merging.

kuzaxak:
By the way, do you have any plans to create integration tests? I see you have some frameworks in dependencies. Perhaps it is worth covering this functionality with tests.

I've added a small set of unit tests, but I am not sure if it is adequate and matches your expectations.

ilya-lesikov:
Tests are good, don't bother. Ultimately I was planning to use ginkgo/gomega for tests, like we do in werf and other places, e.g.: https://github.com/deckhouse/delivery-kit-sdk/blob/main/pkg/signature/elf/inhouse/file_test.go
https://github.com/werf/werf/blob/main/test/e2e/converge/simple_test.go

But that's not a priority now. Nelm is mostly tested in werf for now, since nelm is the werf deployment engine, so if anything breaks I usually see it there.

kuzaxak:
@ilya-lesikov fixed a bug and covered it with a test to validate. Missed that case originally, thanks for checking!

PR #386, opened 2025-07-22
feat(cli): release install, release rollback and release uninstall should have --no-pod-logs option

nervgh:


cmd/nelm/release_install.go
```diff
@@ -275,6 +275,13 @@ func newReleaseInstallCommand(ctx context.Context, afterAllCommandsBuiltFuncs ma
 			return fmt.Errorf("add flag: %w", err)
 		}
 
+		if err := cli.AddFlag(cmd, &cfg.NoLogs, "no-logs", false, "Don't collect logs of deployed pods", cli.AddFlagOptions{
+			GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
+			Group:                miscFlagGroup,
```
ilya-lesikov:
```suggestion
			Group:                progressFlagGroup,
```

cmd/nelm/release_install.go
```diff
@@ -275,6 +275,13 @@ func newReleaseInstallCommand(ctx context.Context, afterAllCommandsBuiltFuncs ma
 			return fmt.Errorf("add flag: %w", err)
 		}
 
+		if err := cli.AddFlag(cmd, &cfg.NoLogs, "no-logs", false, "Don't collect logs of deployed pods", cli.AddFlagOptions{
```
ilya-lesikov:
```suggestion
		if err := cli.AddFlag(cmd, &cfg.NoPodLogs, "no-pod-logs", false, "Disable Pod logs collection and printing", cli.AddFlagOptions{
```

internal/plan/deploy_plan_builder.go
```diff
@@ -865,7 +868,7 @@ func (b *DeployPlanBuilder) setupHookOperations(infos []*info.DeployableHookReso
 					SaveLogsOnlyForContainers:                showLogsOnlyFor,
 					SaveLogsByRegex:                          logRegex,
 					SaveLogsByRegexForContainers:             logRegexesFor,
-					IgnoreLogs:                               info.Resource().SkipLogs(),
+					IgnoreLogs:                               lo.TernaryF(b.ignoreLogs, alwaysTrue, info.Resource().SkipLogs),
```
ilya-lesikov: Please, no. Just compute it somewhere above, e.g. on line 847:
```go
var ignoreLogs bool
if b.ignoreLogs || info.Resource().SkipLogs() {
  ignoreLogs = true
}
```

pkg/action/release_install.go
```diff
@@ -491,27 +493,12 @@ func releaseInstall(ctx context.Context, releaseName, releaseNamespace string, o
 	)
 
 	log.Default.Debug(ctx, "Starting tracking")
-	stdoutTrackerStopCh := make(chan bool)
-	stdoutTrackerFinishedCh := make(chan bool)
+	progressPrinter := newProgressTablePrinter(ctx, opts.ProgressTablePrintInterval, opts.Timeout, func() {
```
ilya-lesikov: opts.Timeout is the global action timeout, which includes more than just tracking. By using it here, you kinda decreased the factual opts.Timeout that users specified previously, since now opts.Timeout is not for the action as a whole, but just for the tracking part.

You could add a global timeout specifically for tracking, but it must be a separate option. I'd say we don't need such a timeout at all, not much value.

ilya-lesikov (changes requested):

ilya-lesikov (approved):

PR #436, opened 2025-10-20
feat: allow to hide insignificant changes details

drey:
Hi,

This PR motivation is to improve user experience with planning. 

Currently, when planning is used, we provide user with information about all the insignificant changes,  that some times makes hard to analyze what significant changes will be made. Presence of sensitive resource changes (marked as ` <hidden sensitive changes>`) make life even harder, because user should go through all the changes and keep an eye on every resource to be sure that it is going to get insignificant changes, but not significant ones (`<hidden insignificant changes>` vs ` <hidden sensitive changes>` vs actual changes for every resource).

Also, currently, if resource is a fully-sensitive resource (`SensitiveInfo.FullySensitive() == true`), it always gets `<hidden sensitive changes>` even if changes actually are insignificant (not related to the secret data).

This PR introduces the following capabilities:

* Add `HideInsignificantChanges` option to `ReleasePlanInstallOptions` that allow not to print details about resources with insignificant changes. Such resources still be added to the total amount of resources changed.
* Mark fully-significant resources by ` <hidden sensitive changes>` only if there are really some significant changes, and mark it by `<hidden insignificant changes>` otherwise (makes possible to hide such changes by above logic)

If you find this PR handly, I would happy to introduce somethings like `--hide-insignificant-changes` to `werf`.

### DIFF
--- a/internal/plan/planned_changes.go
+++ b/internal/plan/planned_changes.go
@@ -40,6 +40,10 @@ type ResourceChange struct {
 	Udiff           string
 }
 
+func (rc *ResourceChange) IsInsignificant() bool {
+	return rc.Udiff == HiddenInsignificantChanges
+}
+
 func CalculatePlannedChanges(installableInfos []*InstallableResourceInfo, deletableInfos []*DeletableResourceInfo, opts CalculatePlannedChangesOptions) ([]*ResourceChange, error) {
 	instInfosByIter := groupInstInfosByIter(installableInfos)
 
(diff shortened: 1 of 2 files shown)


ilya-lesikov:
Doesn't [NELM_FEAT_FIELD_SENSITIVE=true](https://github.com/werf/nelm?tab=readme-ov-file#env-variable-nelm_feat_field_sensitive) help with this? I think it should display actual colored diffs if there are secret changes, while `<hidden insignificat changes>` remain grey. This env is not exposed in werf though, but it can be done. NELM_FEAT_FIELD_SENSITIVE will be true by default in werf v3/nelm v2.

PR #478, opened 2025-11-13
chore(ci): optimize test-pr workflow and enable go caching

odiora:


.github/workflows/test_pr.yml
```diff
@@ -12,47 +12,61 @@ concurrency:
...
+              - '**/*.yml'
+
+      - name: Debug changed files
+        run: |
+          echo "🧩 src_changed=${{ steps.changed-files.outputs.src_any_changed }}"
+          echo "🎨 lint_changed=${{ steps.changed-files.outputs.lint_any_changed }}"
+          echo "SRC files:"
+          echo "${{ steps.changed-files.outputs.src_all_changed_files }}" | tr ' ' '\n'
+          echo "LINT files:"
+          echo "${{ steps.changed-files.outputs.lint_all_changed_files }}" | tr ' ' '\n'
```
Fral738: No need for debug log

.github/workflows/test_pr.yml
```diff
@@ -12,47 +12,61 @@ concurrency:
...
+        uses: tj-actions/changed-files@v47
         with:
-          filters: |
-            workflow_proceed:
+          files_yaml: |
+            src:
+              - '**/*.go'
               - 'go.mod'
+              - 'go.sum'
               - 'Taskfile.dist.yaml'
```
Fral738: Is this file needed here?

.github/workflows/test_pr.yml
```diff
@@ -12,47 +12,61 @@ concurrency:
...
     with:
-      forceSkip: ${{ github.event_name == 'pull_request' && needs.detect-changes.outputs.workflow_proceed == 'false' }}
+      forceSkip: ${{ github.event_name == 'pull_request' && needs.detect-changes.outputs.lint_changed == 'false' }}
 
   unit:
     needs: detect-changes
     uses: ./.github/workflows/_test_unit.yml
     with:
-      forceSkip: ${{ github.event_name == 'pull_request' && needs.detect-changes.outputs.workflow_proceed == 'false' }}
+      forceSkip: ${{ github.event_name == 'pull_request' && needs.detect-changes.outputs.src_changed == 'false' }}
```
Fral738, line 55: Is it worth running tests if the code doesn't change?

odiora: Agree — tests run only if Go sources or dependencies change.
No point wasting CI time when nothing that affects the code was modified.

.github/workflows/_lint.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
```
Fral738, line 31: Use `cache: false` if you use `actions/cache`
https://github.com/actions/setup-go/tree/v6?tab=readme-ov-file#caching-dependency-files-and-build-outputs
> The cache input is optional, and caching is turned on by default.

.github/workflows/_test_unit.yml
```diff
@@ -33,9 +33,8 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
-          cache: true
```
Fral738, line 38: Same

odiora: I’d prefer to keep caching enabled here because unit tests frequently depend on module downloads (go mod download / go test ./...), and built-in caching in setup-go@v6 significantly reduces execution time.

In contrast to the build job — where we use a manual actions/cache@v4 — here there’s no risk of key collisions, so the default cache: true is both safe and efficient.

Fral738 (approved):

PR #497, opened 2025-11-27
docs: add Go templates alternative proposal

ilya-lesikov:


docs/proposals/go-templates-alternative.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Specification
+
+The chart structure with TypeScript support will look like this:
+```
+Chart.yaml
+values.yaml
+templates/
+ts/
+  package.json
```
DmitryBochkarev, line 135: Should we support `main` section for package.json to override entry point?

- https://docs.npmjs.com/cli/v11/configuring-npm/package-json#main

docs/proposals/go-templates-alternative.md
```diff
@@ -0,0 +1,158 @@
...
+  vendor/
+  src/
+    index.ts
+    deployment.ts
+    service.ts
+```
+
+Here the only change is that the new `ts` directory is added. This directory basically represents a TypeScript application. The application must accept Helm root context data (values, chart info, etc.) as input and return rendered manifests as output. The application must be possible to run with `npm start`.
+
+`node_modules` directory must be in .gitignore. During chart publishing all dependencies must be bundled into `vendor/libs.js`, which can be done with esbuild, which is to be embedded in Nelm.
```
DmitryBochkarev: I think there are missing build step of `libs.js`. Which should be commited to git(to make werf-giterminism happy). I'm assume it should be done by manually by running nelm command(something like nelm ts build).

Which is raises a two ralated points:

`libs.js` becomes unreviewable and this can cotradicts org security policies, because build of components should happen on special job pipelines in which automated security check is performed and packages are vetted. This can guarantee that resulted chart is build from packages specified in `package.json`

docs/proposals/go-templates-alternative.md
```diff
@@ -0,0 +1,158 @@
...
+
+Development workflow:
+1. NodeJS, NPM and Nelm must be installed for local development.
+1. Command `nelm chart ts init .` creates `ts` directory with boilerplate files.
+1. `ts` directory opened as a TypeScript NodeJS project in IDE/editor.
+1. Work as you would with a TypeScript project.
+1. Run `npm TODO` to execute application with NodeJS runtime and render manifests to stdout.
+1. Run `npm TODO` to run tests.
+1. Run `nelm chart render` to render manifests with the QuickJS runtime.
+1. Run `nelm chart upload` to publish the chart. Nelm will bundle dependencies into `vendor/libs.js` during publishing with embedded esbuild.
```
DmitryBochkarev: If security policies incorporated, build step is done in ci pipeline. Which means that this pipeline will require npm for packages to be downloaded. Which is contradicts a point that only nelm is needed. Which raises a question will - user expect that nelm(with esbuild) or npm should be responsible for building `libs.js`?

My point that we should be careful with instructions.

Issue #54, opened 2024-07-24
Alternative to Go templates


ilya-lesikov:
### Problem

Go templates which are used to generate Kubernetes manifests in Helm are difficult to write, read, debug and maintain. They have issues with performance, values mutability, template name collisions and others.

### Solution (if you have one)

From what we know, Go templates cannot be improved much. We could add a few template functions to the library, the tooling can be improved, but not much else.

We should consider an alternative to Go templating in Nelm. This can be one of:
* General-purpose language (TypeScript, Python, ...)
* Specialized language (Cue, KCL, ...)
* Templating language

Another templating language won't solve much, so we left with general-purpose languages and specialized languages like Cue.

### Additional information

Helm 3 has `--post-renderer` flag which allows usage of any other language to generate manifests, but it is rarely used due to:
1. Chart developers can't enforce usage of specific post-renderer with a specific configuration for their chart. They must provide instructions to the chart consumers on how to manually set up the post-renderer and how to run it. Go templating works out-of-the box and requires no configuration.
2. Ability to use any language results in chart consumers potentially required to learn many different languages just for Kubernetes manifest generation.

shurup:
> Specialized language 

**KCL** is gaining momentum in the cloud native community right now. It joined CNCF last September only. According to the latest CNCF projects' velocity reports, it had:
* 40 authors and 2632 commits during Jan'2023 — Jan'2024 (106th CNCF project by velocity);
* 72 authors and 2876 commits during Jul'2023 — Jul'2024 (68th).

Can't find an up-to-date list of new projects adopting it, though.

**CUE** is used by Timoni, so we can get some feedback from its users to see whether it's a good fit.

tonnydourado:
-1 for templating, in any shape or form. It's just the worse if both words, you don't get tooling support for the host language NOR for the templating language, most of the time. 

+1 for:
- anything that allow some form of typed interface on charts. It should be easy to see all parameters that a chart expects, with constraints and defaults. The whole `values.yaml` thing is extremely confusing.
- some form of patching. One of the biggest issues I had involved having to change a value that was deep in the chart dependency, but not exposed all the way through.

nicolas-goudry:
[...]

PR #500, opened 2025-11-30
docs: TypeScript charts specification

kitsunoff:
## Summary

- Add detailed specification for TypeScript as alternative to Go templates
- Define HelmContext API and types
- Describe SDK package structure (@nelm/sdk)
- Document development and deployment workflow
- List CLI commands
- Record design decisions

## Files

- `docs/proposals/go-template-alternative/README.md` — Overview
- `docs/proposals/go-template-alternative/api.md` — HelmContext API
- `docs/proposals/go-template-alternative/sdk.md` — SDK structure
- `docs/proposals/go-template-alternative/workflow.md` — Workflow
- `docs/proposals/go-template-alternative/cli.md` — CLI commands
- `docs/proposals/go-template-alternative/decisions.md` — Design decisions

## Key Decisions

1. **Runtime:** quickjs-go (CGO), no WASM
2. **API:** `render(ctx) → Manifest[]` (return-based, not emit)
3. **Context:** Everything via `ctx` (no globals, no imports for runtime)
4. **Subcharts:** Full isolation, Nelm orchestrates
5. **SDK:** Types only + `when()` helper, all runtime from Go
6. **Values types:** npm script with json-schema-to-typescript

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #502, opened 2025-12-06
feat: TypeScript support in charts

DmitryBochkarev:
This PR is based on https://github.com/werf/nelm/pull/497 and https://github.com/werf/nelm/pull/500

Closes https://github.com/werf/nelm/issues/54

I plan to provide more detailed examples in https://github.com/DmitryBochkarev/nelm-ts-examples

This is quite a big PR. I can try to split it into separate commits. 1000 lines of 2000 all implementation code are TS project initialization code(`nelm chart create`), and 2/3 of the rest of the code is tests - so 1,000 lines of actual feature implementation.

This PR is slightly diverged from original proposal regarding target js bundle file name. I named it `ts/chart_render_main.js` so it is explicit and hard to randomly create such file.

TypeScript types a more simple than propose in https://github.com/werf/nelm/pull/500
 but it can be fixed later.

**What is implemented**
- [x] Feature gate NELM_FEAT_TS_CHARTS
- [x] Packaging with transpiling from TypeScript to a JavaScript bundle (one file)
- [x] Rendering packaged or source directory. If a packaged JS bundle is present, rendering uses it instead of trying to transpile.
- [x] SourceMaps with proper backtraces to the original file on render errors
- [x] Generation of a TypeScript project. If the feature is enabled, by default, it initializes a TS project with proper structure. If --only-ts is specified, it adds a TS project to an existing chart or creates a new chart with only TS without a templates directory. Here is generated [chart](https://github.com/DmitryBochkarev/nelm-ts-examples/tree/main/start-chart)
- [x] If no external packages are used there are no external dependencies are needed only nelm

**What is not implemented**
- [ ] Proper documentation for the feature
- [ ] Minification is not implemented because it breaks SourceMaps for some reason (I have not found a relevant issue in esbuild issues)
- [ ] Tree shaking – esbuild documentation states that tree shaking does not work with the CommonJS target
- [ ] ApiVersion for rendered values (only manifests are read), but it is easy to implement
- [ ] Separate npm package for nelm types

**What I need help with**
- I need more complex examples of charts that can push the limits of this feature to eliminate some blind spots
- I need help with proper testing of this feature (develop, pack, upload, download, install)

### DIFF
--- a/cmd/nelm/chart_init.go
+++ b/cmd/nelm/chart_init.go
@@ -1,0 +1,89 @@
+package main
+
+import (
+	"cmp"
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/werf/common-go/pkg/cli"
+	"github.com/werf/nelm/pkg/action"
+	"github.com/werf/nelm/pkg/common"
+	"github.com/werf/nelm/pkg/log"
+)
+
+type chartInitConfig struct {
+	action.ChartInitOptions
+
+	LogColorMode string
+	LogLevel     string
+}
+
+func newChartInitCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*cobra.Command]func(cmd *cobra.Command) error) *cobra.Command {
+	cfg := &chartInitConfig{}
+
+	cmd := cli.NewSubCommand(
+		ctx,
+		"init [PATH]",
+		"Initialize a new chart.",
+		"Initialize a new chart in the specified directory. If PATH is not specified, uses the current directory.",
+		10, // priority for ordering in help
+		chartCmdGroup,
+		cli.SubCommandOptions{
+			Args: cobra.MaximumNArgs(1),
+			ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+				return nil, cobra.ShellCompDirectiveFilterDirs
+			},
+		},
+		func(cmd *cobra.Command, args []string) error {
+			ctx = log.SetupLogging(ctx, cmp.Or(log.Level(cfg.LogLevel), log.InfoLevel), log.SetupLoggingOptions{
+				ColorMode: cfg.LogColorMode,
+			})
+
+			if len(args) > 0 {
+				cfg.ChartDirPath = args[0]
+			}
+
+			if err := action.ChartInit(ctx, cfg.ChartInitOptions); err != nil {
+				return fmt.Errorf("chart init: %w", err)
+			}
+
+			return nil
+		},
+	)
+
+	afterAllCommandsBuiltFuncs[cmd] = func(cmd *cobra.Command) error {
+		if err := cli.AddFlag(cmd, &cfg.TS, "ts", false, "Initialize TypeScript chart", cli.AddFlagOptions{
+			Group: mainFlagGroup,
+		}); err != nil {
+			return fmt.Errorf("add flag: %w", err)
+		}
+
+		if err := cli.AddFlag(cmd, &cfg.TempDirPath, "temp-dir", "", "The directory for temporary files. By default, create a new directory in the default system directory for temporary files", cli.AddFlagOptions{
+			GetEnvVarRegexesFunc: cli.GetFlagGlobalEnvVarRegexes,
+			Group:                miscFlagGroup,
+			Type:                 cli.FlagTypeDir,
+		}); err != nil {
+			return fmt.Errorf("add flag: %w", err)
+		}
+
+		if err := cli.AddFlag(cmd, &cfg.LogColorMode, "color-mode", common.DefaultLogColorMode, "Color mode for logs. "+allowedLogColorModesHelp(), cli.AddFlagOptions{
+			GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
+			Group:                miscFlagGroup,
+		}); err != nil {
+			return fmt.Errorf("add flag: %w", err)
+		}
+
+		if err := cli.AddFlag(cmd, &cfg.LogLevel, "log-level", string(log.InfoLevel), "Set log level. "+allowedLogLevelsHelp(), cli.AddFlagOptions{
+			GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
+			Group:                miscFlagGroup,
+		}); err != nil {
+			return fmt.Errorf("add flag: %w", err)
+		}
+
+		return nil
+	}
+
+	return cmd
+}
--- a/cmd/nelm/chart_pack.go
+++ b/cmd/nelm/chart_pack.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/samber/lo"
@@ -10,6 +11,8 @@ import (
 	helm_v3 "github.com/werf/3p-helm/cmd/helm"
 	"github.com/werf/3p-helm/pkg/chart/loader"
 	"github.com/werf/common-go/pkg/cli"
+	"github.com/werf/nelm/internal/tschart"
+	"github.com/werf/nelm/pkg/featgate"
 	"github.com/werf/nelm/pkg/log"
 )
 
@@ -32,6 +35,15 @@ func newChartPackCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*co
 		ctx = log.SetupLogging(ctx, lo.Ternary(helmSettings.Debug, log.DebugLevel, log.InfoLevel), log.SetupLoggingOptions{})
 
 		loader.NoChartLockWarning = ""
+
+		if featgate.FeatGateTypescript.Enabled() {
+			transformer := tschart.NewTransformer()
+			for _, chartPath := range args {
+				if err := transformer.TransformChartDir(ctx, chartPath); err != nil {
+					return fmt.Errorf("transform TypeScript in %q: %w", chartPath, err)
+				}
+			}
+		}
 
 		if err := originalRunE(cmd, args); err != nil {
 			return err
--- a/internal/tschart/engine.go
+++ b/internal/tschart/engine.go
@@ -1,0 +1,305 @@
+package tschart
+
+import (
+	"context"
+	"fmt"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"sigs.k8s.io/yaml"
+
+	helmchart "github.com/werf/3p-helm/pkg/chart"
+	"github.com/werf/3p-helm/pkg/chartutil"
+	"github.com/werf/nelm/pkg/log"
+)
+
+func mergeRuntimeFiles(runtimeFiles, runtimeDepsFiles []*helmchart.File) []*helmchart.File {
+	if len(runtimeDepsFiles) == 0 {
+		return runtimeFiles
+	}
+
+	merged := make([]*helmchart.File, 0, len(runtimeFiles)+len(runtimeDepsFiles))
+	merged = append(merged, runtimeFiles...)
+	merged = append(merged, runtimeDepsFiles...)
+
+	return merged
+}
+
+type Engine struct{}
+
+func NewEngine() *Engine {
+	return &Engine{}
+}
+
+func (e *Engine) RenderChartWithDependencies(
+	ctx context.Context,
+	rootChartPath string,
+	chart *helmchart.Chart,
+	renderedValues chartutil.Values,
+) (map[string]string, error) {
+	allRendered := make(map[string]string)
+
+	err := e.renderChartRecursive(ctx, rootChartPath, chart, renderedValues, chart.Name(), allRendered)
+	if err != nil {
+		return nil, err
+	}
+
+	return allRendered, nil
+}
+
+//nolint:funcorder // Helper method kept near the method that uses it for readability
+func (e *Engine) renderChartRecursive(
+	ctx context.Context,
+	chartDirPath string, // Filesystem path (for local charts)
+	chart *helmchart.Chart,
+	values chartutil.Values,
+	pathPrefix string, // Output path prefix (e.g., "root/charts/sub")
+	results map[string]string,
+) error {
+	log.Default.Debug(ctx, "Rendering TypeScript for chart %q (path prefix: %s)", chart.Name(), pathPrefix)
+
+	rendered, err := e.RenderFiles(ctx, chartDirPath, chart, values)
+	if err != nil {
+		return fmt.Errorf("render TypeScript for chart %q: %w", chart.Name(), err)
+	}
+
+	for filename, content := range rendered {
+		outputPath := path.Join(pathPrefix, filename)
+		results[outputPath] = content
+		log.Default.Debug(ctx, "Added TypeScript output: %s", outputPath)
+	}
+
+	for _, dep := range chart.Dependencies() {
+		depName := dep.Name()
+
+		log.Default.Debug(ctx, "Processing dependency %q for chart %q", depName, chart.Name())
+
+		depValues := scopeValuesForSubchart(values, depName, dep)
+
+		depDirPath := filepath.Join(chartDirPath, "charts", depName)
+
+		depPathPrefix := path.Join(pathPrefix, "charts", depName)
+
+		err := e.renderChartRecursive(ctx, depDirPath, dep, depValues, depPathPrefix, results)
+		if err != nil {
+			return fmt.Errorf("render dependency %q: %w", depName, err)
+		}
+	}
+
+	return nil
+}
+
+func scopeValuesForSubchart(parentValues chartutil.Values, subchartName string, subchart *helmchart.Chart) chartutil.Values {
+	scoped := chartutil.Values{}
+
+	if caps, ok := parentValues["Capabilities"]; ok {
+		scoped["Capabilities"] = caps
+	}
+
+	if release, ok := parentValues["Release"]; ok {
+		scoped["Release"] = release
+	}
+
+	if runtime, ok := parentValues["Runtime"]; ok {
+		scoped["Runtime"] = runtime
+	}
+
+	scoped["Chart"] = buildChartMetadata(subchart)
+
+	if parentVals, ok := parentValues["Values"]; ok {
+		switch v := parentVals.(type) {
+		case map[string]interface{}:
+			if subVals, ok := v[subchartName]; ok {
+				scoped["Values"] = subVals
+			} else {
+				scoped["Values"] = map[string]interface{}{}
+			}
+		case chartutil.Values:
+			if subVals, ok := v[subchartName]; ok {
+				scoped["Values"] = subVals
+			} else {
+				scoped["Values"] = map[string]interface{}{}
+			}
+		default:
+			scoped["Values"] = map[string]interface{}{}
+		}
+	} else {
+		scoped["Values"] = map[string]interface{}{}
+	}
+
+	files := make(map[string]interface{}, len(subchart.Files))
+	for _, file := range subchart.Files {
+		files[file.Name] = file.Data
+	}
+
+	scoped["Files"] = files
+
+	return scoped
+}
+
+func buildChartMetadata(chart *helmchart.Chart) map[string]interface{} {
+	metadata := map[string]interface{}{
+		"Name":    chart.Name(),
+		"Version": "",
+	}
+
+	if chart.Metadata != nil {
+		metadata["Version"] = chart.Metadata.Version
+		metadata["AppVersion"] = chart.Metadata.AppVersion
+		metadata["Description"] = chart.Metadata.Description
+		metadata["Keywords"] = chart.Metadata.Keywords
+		metadata["Home"] = chart.Metadata.Home
+		metadata["Sources"] = chart.Metadata.Sources
+		metadata["Icon"] = chart.Metadata.Icon
+		metadata["APIVersion"] = chart.Metadata.APIVersion
+		metadata["Condition"] = chart.Metadata.Condition
+		metadata["Tags"] = chart.Metadata.Tags
+		metadata["Type"] = chart.Metadata.Type
+		metadata["Annotations"] = chart.Metadata.Annotations
+
+		if chart.Metadata.Maintainers != nil {
+			maintainers := make([]map[string]interface{}, len(chart.Metadata.Maintainers))
+			for i, m := range chart.Metadata.Maintainers {
+				maintainers[i] = map[string]interface{}{
+					"Name":  m.Name,
+					"Email": m.Email,
+					"URL":   m.URL,
+				}
+			}
+
+			metadata["Maintainers"] = maintainers
+		}
+	}
+
+	return metadata
+}
+
+func (e *Engine) RenderFiles(ctx context.Context, chartPath string, chart *helmchart.Chart, renderedValues chartutil.Values) (map[string]string, error) {
+	mergedFiles := mergeRuntimeFiles(chart.RuntimeFiles, chart.RuntimeDepsFiles)
+
+	var (
+		vendorBundle string
+		packages     []string
+		entrypoint   string
+	)
+
+	vendorBundle, packages, err := GetVendorBundleFromFiles(mergedFiles)
+	if err != nil {
+		return nil, fmt.Errorf("get vendor bundle: %w", err)
+	}
+
+	sourceFiles := ExtractSourceFiles(mergedFiles)
+	if len(sourceFiles) == 0 {
+		return map[string]string{}, nil
+	}
+
+	entrypoint = findEntrypointFromFiles(sourceFiles)
+	if entrypoint == "" {
+		return map[string]string{}, nil
+	}
+
+	appBundle, err := BuildAppBundleFromChartFiles(ctx, mergedFiles, packages)
+	if err != nil {
+		return nil, fmt.Errorf("build app bundle from chart files: %w", err)
+	}
+
+	renderContext := buildRenderContext(renderedValues, chart)
+
+	result, err := executeInGoja(vendorBundle, appBundle, renderContext)
+	if err != nil {
+		return nil, fmt.Errorf("execute bundle: %w", err)
+	}
+
+	if result == nil {
+		return map[string]string{}, nil
+	}
+
+	yamlOutput, err := resultToYAML(result)
+	if err != nil {
+		return nil, fmt.Errorf("convert result to YAML: %w", err)
+	}
+
+	if strings.TrimSpace(yamlOutput) == "" {
+		return map[string]string{}, nil
+	}
+
+	outputPath := path.Join(TSSourceDir, entrypoint)
+
+	return map[string]string{
+		outputPath: yamlOutput,
+	}, nil
+}
+
+func findEntrypointFromFiles(sourceFiles map[string][]byte) string {
+	for _, ep := range EntryPoints {
+		if _, exists := sourceFiles[ep]; exists {
+			return ep
+		}
+	}
+
+	return ""
+}
+
+func buildRenderContext(renderedValues chartutil.Values, chart *helmchart.Chart) map[string]interface{} {
+	renderContext := renderedValues.AsMap()
+
+	if valuesInterface, ok := renderContext["Values"]; ok {
+		if chartValues, ok := valuesInterface.(chartutil.Values); ok {
+			renderContext["Values"] = chartValues.AsMap()
+		}
+	}
+
+	renderContext["Chart"] = buildChartMetadata(chart)
+
+	files := make(map[string]interface{}, len(chart.Files))
+	for _, file := range chart.Files {
+		files[file.Name] = file.Data
+	}
+
+	renderContext["Files"] = files
+
+	return renderContext
+}
+
+func resultToYAML(result interface{}) (string, error) {
+	resultMap, ok := result.(map[string]interface{})
+	if !ok {
+		return "", fmt.Errorf("unexpected render result type: %T", result)
+	}
+
+	manifests, exists := resultMap["manifests"]
+	if !exists {
+		return "", fmt.Errorf("render result object does not contain 'manifests' field")
+	}
+
+	return convertToYAML(manifests)
+}
+
+func convertToYAML(value interface{}) (string, error) {
+	if arr, ok := value.([]interface{}); ok {
+		var results []string
+
+		for i, item := range arr {
+			if item == nil {
+				continue
+			}
+
+			yamlBytes, err := yaml.Marshal(item)
+			if err != nil {
+				return "", fmt.Errorf("marshal resource at index %d: %w", i, err)
+			}
+
+			results = append(results, string(yamlBytes))
+		}
+
+		return strings.Join(results, "---\n"), nil
+	}
+
+	yamlBytes, err := yaml.Marshal(value)
+	if err != nil {
+		return "", fmt.Errorf("marshal resource: %w", err)
+	}
+
+	return string(yamlBytes), nil
+}
--- a/internal/tschart/init.go
+++ b/internal/tschart/init.go
@@ -1,0 +1,396 @@
+package tschart
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/werf/nelm/pkg/log"
+)
+
+// InitChartStructure creates Chart.yaml and values.yaml if they don't exist.
+// For .helmignore: creates if missing, or appends TS entries if exists.
+// Returns error if ts/ directory already exists.
+func InitChartStructure(ctx context.Context, chartPath, chartName string) error {
+	tsDir := filepath.Join(chartPath, TSSourceDir)
+	if _, err := os.Stat(tsDir); err == nil {
+		return fmt.Errorf("TypeScript directory already exists: %s. Cannot initialize in a directory with existing TypeScript chart files", tsDir)
+	} else if !os.IsNotExist(err) {
+		return fmt.Errorf("stat %s: %w", tsDir, err)
+	}
+
+	skipIfExists := []struct {
+		path    string
+		content string
+	}{
+		{filepath.Join(chartPath, "Chart.yaml"), generateChartYaml(chartName)},
+		{filepath.Join(chartPath, "values.yaml"), generateValuesYaml()},
+	}
+
+	for _, f := range skipIfExists {
+		_, err := os.Stat(f.path)
+		if err != nil && !os.IsNotExist(err) {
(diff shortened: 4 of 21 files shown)


cmd/nelm/chart_create.go
```diff
@@ -0,0 +1,126 @@
...
+	cli.SetSubCommandAnnotations(cmd, 10, chartCmdGroup)
+
+	originalRunE := cmd.RunE
+	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		helmSettings := helm_v3.Settings
+		ctx = log.SetupLogging(ctx, lo.Ternary(helmSettings.Debug, log.DebugLevel, log.InfoLevel), log.SetupLoggingOptions{})
+
+		// Handle --only-ts mode
+		if cfg.OnlyTS {
+			if !featgate.FeatGateTSCharts.Enabled() {
```
ilya-lesikov: I guess we shouldn't add the `nelm chart create` command at all if FeatGateTSCharts is not enabled.

cmd/nelm/chart_create.go
```diff
@@ -0,0 +1,126 @@
...
+			// Check if Chart.yaml exists
+			chartYamlPath := filepath.Join(chartPath, "Chart.yaml")
+			chartExists := false
+			if _, err := os.Stat(chartYamlPath); err == nil {
+				chartExists = true
+			}
+
+			// If no existing chart, create chart structure (without templates/)
+			if !chartExists {
+				if err := tschart.CreateTSOnlyChartStructure(ctx, chartPath, chartName); err != nil {
```
ilya-lesikov: I see we do a lot of stuff in CreateTSOnlyChartStructure. I was thinking only about populating `ts` directory. But maybe it's not feasible (we still need to update at least `.helmignore`). So let's rename the option `--ts-only` to `--ts`

cmd/nelm/chart_create.go
```diff
@@ -0,0 +1,126 @@
...
+
+func newChartCreateCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*cobra.Command]func(cmd *cobra.Command) error) *cobra.Command {
+	cfg := &chartCreateConfig{}
+
+	cmd := lo.Must(lo.Find(helmRootCmd.Commands(), func(c *cobra.Command) bool {
+		return strings.HasPrefix(c.Use, "create")
+	}))
+
+	cmd.LocalFlags().AddFlagSet(cmd.InheritedFlags())
+	cmd.Use = "create NAME [PATH]"
```
ilya-lesikov: I wanted it to be `init`, not `create`, because I think it's less about creating a chart, and more about initializing it (or some parts of it). E. g. one of the main use cases for the command will be initializing `ts` dir to work with TS, which is not exactly "creating" a chart.

cmd/nelm/chart_create.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+type chartCreateConfig struct {
+	OnlyTS bool
+}
+
+func newChartCreateCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*cobra.Command]func(cmd *cobra.Command) error) *cobra.Command {
+	cfg := &chartCreateConfig{}
+
+	cmd := lo.Must(lo.Find(helmRootCmd.Commands(), func(c *cobra.Command) bool {
```
ilya-lesikov: Just realised, you tried to use `helm create` command. Let's not do this. Just generate `.helmignore` and `ts` and `Chart.yaml` with apiVersion, name, version (three required fields) and `values.yaml` with like a few keys. That's gonna be enough for now. Fail if `--ts` flag not specified with something like "not implemented yet" error.

ilya-lesikov: And since we don't use `helm create` command, implement this command like we do all the other native commands, like `chart lint`: https://github.com/werf/nelm/blob/main/cmd/nelm/chart_lint.go

internal/tschart/transformer.go
```diff
@@ -0,0 +1,259 @@
...
+// - If bundle already exists in chart.Files (loaded by Helm or pre-built), returns nil
+// - If local directory with entrypoint: builds from filesystem, adds to chart.Files
+// - If packaged/remote chart with source but no bundle: returns error
+//
+// Unlike TransformChartDir (which writes to disk for packaging), this function
+// only adds the bundle to chart.Files for in-memory rendering.
+func (t *Transformer) TransformChartForRender(ctx context.Context, chartPath string, chart *helmchart.Chart) error {
+	// Check if bundle already exists in chart.Files
+	// (either loaded by Helm from disk, or pre-built in packaged chart)
+	for _, f := range chart.Files {
```
ilya-lesikov: I just realized. We need to filter both ts/node_modules and BundleFile out, removing them from chart.Files. Otherwise they are gonna be saved in the Helm Release Secret, because the whole Chart structure is saved in here. We can't do that, because the maximum Kubernetes Secret size is like ~1MB in general. So we need to remove them from chart.Files and all the .Files of the dependent charts.

I guess we can save them to a new field chart.RuntimeFiles instead:
```go
	RuntimeFiles []*File `json:"-"`
```

This way it's not gonna pollute the Secret resource, because it's gonna be dropped on json marshalling.

Just add the new `case` statement [here](https://github.com/werf/3p-helm/blob/17dd24e6c20b82cf6499935140fc700a806f29a3/pkg/chart/loader/load.go#L176) to do this.

internal/tschart/transformer.go
```diff
@@ -0,0 +1,259 @@
...
+//
+// Behavior:
+// - If chartPath is not a directory, returns nil (no-op)
+// - If no ts/src/index.ts or ts/src/index.js exists, returns nil (no-op)
+// - If TypeScript source exists, always overwrites existing bundle
+//
+// Returns error if:
+// - TypeScript files exist but transpilation fails
+// - Cannot write output file
+func (t *Transformer) TransformChartDir(ctx context.Context, chartPath string) error {
```
ilya-lesikov: I thought we are going to generate vendor/libs.js or something like that, not chart_render_main.js. Can't we generate and publish vendor/libs.js on `nelm chart pack/upload`, then on `nelm release install` do this:
1. Transpile all src/**/*.ts to JS in-memory, no changes to fs or chart
2. Feed resulting JS along with vendor/libs.js to goja

I want use to be able to do `helm pull --untar`, then change any .ts files, then do `nelm release install` and it should work (nothing in the filesystem should be mutated, that's important for giterminism in werf).

cmd/nelm/chart_pack.go
```diff
@@ -33,6 +36,15 @@ func newChartPackCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*co
 
 		loader.NoChartLockWarning = ""
 
+		if featgate.FeatGateTSCharts.Enabled() {
```
ilya-lesikov: The same change should be done to `chart_upload.go`

internal/tschart/engine.go
```diff
@@ -0,0 +1,174 @@
...
+	return renderContext
+}
+
+func (e *Engine) executeBundle(ctx context.Context, bundleContent []byte, renderCtx map[string]interface{}) (string, error) {
+	vm, requireModule, err := e.createVM(ctx, bundleContent)
+	if err != nil {
+		return "", fmt.Errorf("create JS VM: %w", err)
+	}
+
+	timeoutCtx, cancel := context.WithTimeout(ctx, e.timeout)
```
ilya-lesikov: 3 seconds timeout? That's kinda rough. Let's get rid of this timeout at all, for now.

internal/tschart/helpers/helpers.go
```diff
@@ -0,0 +1,185 @@
...
+	"crypto/sha256"
+	"encoding/base64"
+	"fmt"
+	"unicode/utf8"
+
+	"github.com/dop251/goja"
+	"github.com/dop251/goja_nodejs/require"
+)
+
+func RegisterHelpersModule(registry *require.Registry) {
```
ilya-lesikov: I don't want any helpers here. Maybe that's something temporary, idk?

Overall I want to be goja runtime be as vanilla as possible. All kinds of functions should be either in Nelm SDK npm package or in third-party js modules.

And at this point I only want to have types in Nelm SDK, so if the user needs base64 functions or something like that, he should go and use some third-party library.

The goal is to make it possible to develop under Node.js, but do `release install` under goja, with as little issues as possible. Helpers embedded in goja will surely complicate development with Node.js. Also, don't see a good reason to force any kind of helpers via Nelm SDK, when there are so much third-party libraries.

internal/tschart/create.go
```diff
@@ -0,0 +1,1063 @@
...
+
+	files := map[string]string{
+		filepath.Join(srcDir, "index.ts"):     generateIndexTS(),
+		filepath.Join(srcDir, "helpers.ts"):   generateHelpersTS(),
+		filepath.Join(srcDir, "resources.ts"): generateResourcesTS(),
+		filepath.Join(typesDir, "nelm.d.ts"):  generateNelmDTS(),
+		filepath.Join(tsDir, "tsconfig.json"): generateTSConfig(),
+		filepath.Join(tsDir, "package.json"):  generatePackageJSON(chartName),
+		filepath.Join(tsDir, ".gitignore"):    generateTSGitignore(),
+	}
```
ilya-lesikov: If any file to be created exists, then we should fail I think, and say that we aren't gonna overwrite this.

ilya-lesikov: .helmignore is an exception, let's append it, if it exists. It's way too imporant.

internal/tschart/create.go
```diff
@@ -0,0 +1,1063 @@
...
+}
+
+func generateIndexTS() string {
+	return `import { RenderContext, RenderResult } from '../types/nelm';
+import {
+  createDeployment,
+  createService,
+  createServiceAccount,
+  createIngress,
+  createHPA,
```
ilya-lesikov: I think we should simplify our demo. In practice people usually remove most of this, and do it themselves. Two resources are going to be fine: deployment and service.

ilya-lesikov: Service can be optional.

internal/tschart/create.go
```diff
@@ -0,0 +1,1063 @@
...
+ */
+export function trunc(str: string, max: number): string {
+  if (str.length <= max) return str;
+  return str.slice(0, max).replace(/-+$/, '');
+}
+
+/**
+ * Get the chart name, respecting nameOverride.
+ */
+export function name(context: RenderContext): string {
```
ilya-lesikov: let's add `get` to all of these, like `getName`

internal/tschart/create.go
```diff
@@ -0,0 +1,1063 @@
...
+  const override = context.Values.nameOverride;
+  return trunc(override || context.Chart.Name, 63);
+}
+
+/**
+ * Get the fully qualified app name.
+ * Truncated at 63 chars (DNS naming spec limit).
+ */
+export function fullname(context: RenderContext): string {
```
ilya-lesikov: let's have `$` instead of `context`, everywhere

internal/tschart/create.go
```diff
@@ -0,0 +1,1063 @@
...
+}
+
+func generateResourcesTS() string {
+	return `import { RenderContext } from '../types/nelm';
+import { fullname, labels, selectorLabels, serviceAccountName } from './helpers';
+
+/**
+ * Create a Deployment resource.
+ */
+export function createDeployment(context: RenderContext): object {
```
ilya-lesikov:
```suggestion
export function newDeployment(context: RenderContext): object {
```

internal/tschart/create.go
```diff
@@ -0,0 +1,1063 @@
...
+    apiVersion: 'apps/v1',
+    kind: 'Deployment',
+    metadata: {
+      name,
+      labels: labels(context),
+    },
+    spec: {
+      ...(Values.autoscaling?.enabled
+        ? {}
+        : { replicas: Values.replicaCount ?? 1 }),
```
ilya-lesikov: Looks scary, can we do these like that? (not sure about null or undefined, we need to either drop it or make it `replicas: null` on marshalling)

```suggestion
      replicas: Values.autoscaling?.enabled ? null : 1,
```

internal/tschart/create.go
```diff
@@ -0,0 +1,1063 @@
...
+node_modules/
+
+# TypeScript build output
+dist/
+`
+}
+
+func generateChartYaml(chartName string) string {
+	return fmt.Sprintf(`apiVersion: v2
+name: %s
```
ilya-lesikov: Let's make it minimal. name, apiVersion, version (three required fields). No comments

internal/tschart/create.go
```diff
@@ -0,0 +1,1063 @@
...
+}
+
+func generateValuesYaml(chartName string) string {
+	return fmt.Sprintf(`# Default values for %s.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
```
ilya-lesikov: Let's make it minimal too. 5-10 values.

internal/tschart/create.go
```diff
@@ -0,0 +1,1063 @@
...
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
```
ilya-lesikov: That's responsibility of the user.
```suggestion
```

pkg/featgate/feat.go
```diff
@@ -47,6 +47,11 @@ var (
 		"clean-null-fields",
 		`Enable cleaning of null fields from resource manifests for better Helm chart compatibility`,
 	)
+
+	FeatGateTSCharts = NewFeatGate(
+		"ts-charts",
```
ilya-lesikov: let's name it `typescript`

internal/tschart/transformer.go
DmitryBochkarev: PR created https://github.com/werf/3p-helm/pull/376

cmd/nelm/chart_init.go
```diff
@@ -0,0 +1,101 @@
...
+			if err := tschart.EnsureGitignore(absPath); err != nil {
+				return fmt.Errorf("ensure .gitignore: %w", err)
+			}
+
+			log.Default.Info(ctx, "Initialized TypeScript chart in %s", absPath)
+			return nil
+		},
+	)
+
+	afterAllCommandsBuiltFuncs[cmd] = func(cmd *cobra.Command) error {
```
ilya-lesikov, line 56: LogLevel LogColorMode TempDirPath are missing

cmd/nelm/chart_init.go
```diff
@@ -0,0 +1,101 @@
...
+		func(cmd *cobra.Command, args []string) error {
+			ctx = log.SetupLogging(ctx, log.InfoLevel, log.SetupLoggingOptions{})
+
+			// Determine chart path
+			chartPath := "."
+			if len(args) > 0 {
+				chartPath = args[0]
+			}
+
+			// Convert to absolute path for reliable name extraction
```
ilya-lesikov: Move the code from here to below to a separate action

cmd/nelm/chart_init.go
```diff
@@ -0,0 +1,101 @@
...
+		10, // priority for ordering in help
+		chartCmdGroup,
+		cli.SubCommandOptions{
+			Args: cobra.MaximumNArgs(1),
+			ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+				return nil, cobra.ShellCompDirectiveFilterDirs
+			},
+		},
+		func(cmd *cobra.Command, args []string) error {
+			ctx = log.SetupLogging(ctx, log.InfoLevel, log.SetupLoggingOptions{})
```
ilya-lesikov: Do it like here (except don't set LogIsParseable):
https://github.com/werf/nelm/blob/675f5d92b5046735d0c0c43d229cc23873d4d074/cmd/nelm/chart_render.go#L48-L51

internal/tschart/transformer.go
```diff
@@ -99,19 +381,16 @@ func formatBuildErrors(errors []esbuild.Message) error {
...
-//
-// Returns error if:
-// - TypeScript files exist but transpilation fails
-// - Cannot write output file
+func findEntrypoint(tsDir string) string {
+	for _, ep := range EntryPoints {
+		epPath := filepath.Join(tsDir, ep)
+		if _, err := os.Stat(epPath); err == nil {
+			return ep
+		}
```
ilya-lesikov: You should hard fail on any unexpected error, like permission denied. "" should be returned only if there is no file.

internal/tschart/transformer.go
```diff
@@ -13,49 +16,329 @@ import (
...
-	BundleFile = "ts/chart_render_main.js"
-	TSConfigFile = "tsconfig.json"
+	TSSourceDir      = "ts/"
+	VendorBundleFile = "ts/vendor/libs.js"
+	VendorBundleDir  = "ts/vendor"
+	TSConfigFile     = "tsconfig.json"
 )
 
 var (
 	EntryPoints = []string{"src/index.ts", "src/index.js"}
```
ilya-lesikov: Can't we reliably get it from package.json or from somewhere? If not, don't bother

internal/tschart/transformer.go
ilya-lesikov: I see you use os.Stat in similar fashion in many other places. Redo it there too.

internal/tschart/init.go
```diff
@@ -0,0 +1,414 @@
...
+	content := string(existingContent)
+	if strings.Contains(content, "ts/node_modules/") {
+		return nil
+	}
+
+	tsEntries := `
+# TypeScript chart files
+ts/node_modules/
+ts/dist/
+`
```
ilya-lesikov, line 129: Reuse, already defined in another place

internal/tschart/init.go
```diff
@@ -0,0 +1,414 @@
...
+	srcDir := filepath.Join(tsDir, "src")
+	typesDir := filepath.Join(tsDir, "types")
+
+	files := []struct {
+		path    string
+		content string
+	}{
+		{filepath.Join(srcDir, "index.ts"), generateIndexTS()},
+		{filepath.Join(srcDir, "helpers.ts"), generateHelpersTS()},
+		{filepath.Join(srcDir, "resources.ts"), generateResourcesTS()},
```
ilya-lesikov: Can we pls split this into deployment.ts, service.ts? Like people usually do with helm templates

internal/tschart/init.go
```diff
@@ -0,0 +1,414 @@
...
+		{filepath.Join(typesDir, "nelm.d.ts"), generateNelmDTS()},
```
ilya-lesikov: I guess that's temporary... Or it shouldn't be? Since we only want types in nelm SDK, at least for now, maybe it's gonna be easier to just generate the types and put them here...

The issue is how we gonna manage backwards compatibility. If some field is added to, let's say, Capabilities, what we do?

internal/tschart/init.go
```diff
@@ -0,0 +1,414 @@
...
+		{filepath.Join(tsDir, "package.json"), generatePackageJSON(chartName)},
+		{filepath.Join(tsDir, ".gitignore"), generateTSGitignore()},
+	}
+
+	for _, f := range files {
+		if _, err := os.Stat(f.path); err == nil {
+			return fmt.Errorf("TypeScript file already exists: %s. Cannot initialize in a directory with existing TypeScript chart files", f.path)
+		}
+	}
+
```
ilya-lesikov: Let's just check if `ts` directory exists. If it does, then fail

internal/tschart/init.go
```diff
@@ -0,0 +1,414 @@
...
+		path    string
+		content string
+	}{
+		{filepath.Join(srcDir, "index.ts"), generateIndexTS()},
+		{filepath.Join(srcDir, "helpers.ts"), generateHelpersTS()},
+		{filepath.Join(srcDir, "resources.ts"), generateResourcesTS()},
+		{filepath.Join(typesDir, "nelm.d.ts"), generateNelmDTS()},
+		{filepath.Join(tsDir, "tsconfig.json"), generateTSConfig()},
+		{filepath.Join(tsDir, "package.json"), generatePackageJSON(chartName)},
+		{filepath.Join(tsDir, ".gitignore"), generateTSGitignore()},
```
ilya-lesikov: Can we move it a level up? To the chart root. Handle it like we do with .helmignore

ilya-lesikov: Wait, this already happens in tschart.EnsureGitignore(). We probably should just remove this line from here.

cmd/nelm/chart_init.go
```diff
@@ -0,0 +1,101 @@
...
+				return fmt.Errorf("ensure .gitignore: %w", err)
+			}
+
+			log.Default.Info(ctx, "Initialized TypeScript chart in %s", absPath)
+			return nil
+		},
+	)
+
+	afterAllCommandsBuiltFuncs[cmd] = func(cmd *cobra.Command) error {
+		if err := cli.AddFlag(cmd, &cfg.TS, "ts", false, "Initialize TypeScript chart (requires NELM_FEAT_TYPESCRIPT=true)", cli.AddFlagOptions{
```
ilya-lesikov:
```suggestion
		if err := cli.AddFlag(cmd, &cfg.TS, "ts", false, "Initialize TypeScript chart", cli.AddFlagOptions{
```

cmd/nelm/chart_init.go
ilya-lesikov: Also, move cfg.TS in action Options too

internal/tschart/transformer.go
```diff
@@ -0,0 +1,585 @@
...
+	"sort"
+	"strings"
+
+	esbuild "github.com/evanw/esbuild/pkg/api"
+	helmchart "github.com/werf/3p-helm/pkg/chart"
+	"github.com/werf/nelm/pkg/log"
+)
+
+const (
+	TSSourceDir      = "ts/"
```
ilya-lesikov, line 15: Having this as a constant is good. But would be great to use this constant in every place where `ts` dir is specified: in many places it is still hardcoded

internal/tschart/transformer.go
```diff
@@ -0,0 +1,585 @@
...
+
+	tsDir := filepath.Join(chartPath, TSSourceDir)
+	if _, err := os.Stat(tsDir); os.IsNotExist(err) {
+		log.Default.Debug(ctx, "No %s directory found, skipping transformation", TSSourceDir)
+		return nil
+	}
+
+	entrypointFile := findEntrypoint(tsDir)
+	if entrypointFile == "" {
+		log.Default.Debug(ctx, "No TypeScript entrypoint found, skipping transformation")
```
ilya-lesikov: Let's fail here instead

internal/tschart/transformer.go
```diff
@@ -0,0 +1,585 @@
...
+		return nil
+	}
+
+	nodeModulesPath := filepath.Join(tsDir, "node_modules")
+	if _, err := os.Stat(nodeModulesPath); os.IsNotExist(err) {
+		log.Default.Debug(ctx, "No node_modules directory found, skipping vendor bundle")
```
ilya-lesikov: We should handle these cases:
1. node_modules are up-to-date: proceed
2. node_modules are not-up-to-date: error
3. node_modules are empty/nonexistent, but there are dependencies in the project: error
4. node_modules are empty/nonexistent, but there are no dependencies in the project: proceed
5. node_modules are not empty, but there are no dependencies: error

Not sure how can we do it. And can we do this reliably and programmatically at all.

internal/tschart/transformer.go
```diff
@@ -0,0 +1,585 @@
...
+	if err := os.WriteFile(vendorPath, []byte(vendorBundle), 0644); err != nil {
+		return fmt.Errorf("write vendor bundle to %s: %w", vendorPath, err)
+	}
+
+	log.Default.Info(ctx, "Wrote vendor bundle to %s", VendorBundleFile)
+
+	return nil
+}
+
+func (t *Transformer) TransformChartForRender(ctx context.Context, chartPath string, chart *helmchart.Chart) error {
```
ilya-lesikov: I guess this whole function is useless at this point? There is no transformation, just a few debug statements. I guess you should merge this code with some other function called later (if there is anything useful left here).

ilya-lesikov: Although it would be good to somehow detect broken ts chart and fail. What can we check? Maybe error if:
1. There are deps, but node_modules and vendor/libs.js are non-existent/empty
2. No entrypoint file

Also, if we validate it once early, we don't need to constantly check for deps and entrypoint every time.

internal/tschart/transformer.go
```diff
@@ -0,0 +1,585 @@
...
+	tsDir := filepath.Join(absChartPath, TSSourceDir)
+	nodeModulesPath := filepath.Join(tsDir, "node_modules")
+	vendorPath := filepath.Join(absChartPath, VendorBundleFile)
+
+	entrypointFile := findEntrypoint(tsDir)
+	if entrypointFile == "" {
+		return "", nil, nil
+	}
+
+	if _, err := os.Stat(nodeModulesPath); err == nil {
```
ilya-lesikov: Can we also check for its emptiness, and handle it like it doesn't exist if empty?

internal/tschart/engine.go
```diff
@@ -0,0 +1,156 @@
...
+
+func (e *Engine) RenderFiles(ctx context.Context, chartPath string, chart *helmchart.Chart, renderedValues chartutil.Values) (map[string]string, error) {
+	isLocalDir := isLocalDirectory(chartPath)
+
+	var vendorBundle string
+	var packages []string
+	var appBundle string
+	var err error
+
+	if isLocalDir {
```
ilya-lesikov: I'm not sure I understood what we are doing here.

chart.RuntimeFiles (and maybe some in chart.Files?) should be populated with all we need, we don't need to go to filesystem for this. You don't wanna work with FS, believe me: remote vs local charts, dependencies in chart tree as archives, as directories, or cached somewhere in `.werf`, etc. It's complex, and all of this is handled in load.Load(), where we build the Chart{}. So just use `chart` struct.


Also, we need to handle `ts` from subcharts in the same manner, are we doing it? I don't see where.

The logic should be like this: get all the files you need from chart.RuntimeFiles and render `ts` for the chart, then go to the next chart via chart.Dependencies() and do the same thing for it. At the end concatenate all results from all the charts/subcharts.

internal/tschart/engine.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return false
+	}
+	return stat.IsDir()
+}
+
+func buildRenderContext(renderedValues chartutil.Values, chart *helmchart.Chart) map[string]interface{} {
+	renderContext := renderedValues.AsMap()
+
+	if valuesInterface, ok := renderContext["Values"]; ok {
```
ilya-lesikov, line 247: Some things are missing: Capabilities, Chart, Release, Runtime (this one is nelm-specific). They were already computed somewhere, we just need to pass them here

internal/tschart/engine.go
```diff
@@ -0,0 +1,156 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	helmchart "github.com/werf/3p-helm/pkg/chart"
+	"github.com/werf/3p-helm/pkg/chartutil"
+	"sigs.k8s.io/yaml"
+)
+
+const OutputFile = "ts/render_output.yaml"
```
ilya-lesikov: Should be something like this:
```
<chart-path>/ts/<path-to-entrypoint>
```

Where chart-path is dynamically computed, and can be something like `rootchart/subchart/subchart` (you can get this info from Chart{} and Chart{}.Dependencies()).

path-to-entrypoint should also be dynamically computed, like we do in all the other places with entrypoint.

internal/tschart/helpers/console.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/dop251/goja"
+
+	"github.com/werf/nelm/pkg/log"
+)
+
+func SetupConsoleGlobal(ctx context.Context, runtime *goja.Runtime) {
```
ilya-lesikov: Hey, how about we use something like github.com/dop251/goja_nodejs/ instead? I really don't want to maintain something like this by myself. Inject whatever you seem useful from goja_nodejs lib into goja runtime, so it behaves like nodejs?

internal/tschart/transformer.go
```diff
@@ -0,0 +1,585 @@
...
+	fileExists := func(path string) bool {
+		_, exists := files[path]
+		return exists
+	}
+
+	tryResolve := func(basePath string) string {
+		if fileExists(basePath) {
+			return basePath
+		}
+		for _, ext := range []string{".ts", ".tsx", ".js", ".jsx", ".json"} {
```
ilya-lesikov: Here and in a few other places I see these hardcoded extensions. What are we doing here with this virtual-fs?

internal/tschart/init.go
DmitryBochkarev: I think there are should be a separate npm-package for nelm and we can validate in nelm that version is compatible with currently used package. Or if we support backwards compatibility, we can allow user to use outdated types definitions, if user wants to have have actual types it can update package from npm.

Another path - is keep this as it allow to regenerate this file by some command.

I think we should implement as proper package

internal/tschart/transformer.go
DmitryBochkarev: We have 2 places where src can be defined:

- for typescript we can specify only source files https://www.typescriptlang.org/tsconfig/#include. We can try to guest from source files what should be picked
- in package.json there are sections that can describe entry points, but only for already compiled files or for projects that are plain javascript

I propose to keep current approach and document it

ilya-lesikov: Alright, let's keep it.

internal/tschart/init.go
ilya-lesikov: I guess you are right, let's go with the separate package. And if we provide some helpers or abstractions libraries down the line, we would still need to make separate packages for this.

internal/tschart/engine.go
DmitryBochkarev: they are present i tested it with:

```ts
import { RenderContext, RenderResult } from '../types/nelm';

export function render($: RenderContext): RenderResult {
  const debugInfo = {
    apiVersion: 'v1',
    kind: 'ConfigMap',
    metadata: {
      name: 'render-context-debug',
    },
    data: {
      'values': JSON.stringify($.Values, null, 2),
      'release': JSON.stringify($.Release, null, 2),
      'chart': JSON.stringify($.Chart, null, 2),
      'capabilities': JSON.stringify($.Capabilities, null, 2),
      'runtime': JSON.stringify(($ as any).Runtime, null, 2),
      'files-keys': JSON.stringify(Object.keys($.Files || {}), null, 2),
    },
  };
  return { manifests: [debugInfo] };
}
```

here is output(`NELM_FEAT_TYPESCRIPT=true ./bin/nelm chart render --log-level=debug  ../nelm-ts-charts/ts-test-chart/ --set-runtime-json=a=1`):

```yaml
---
# Source: ts-test-chart/ts/src/index.ts
apiVersion: v1
data:
  capabilities: |-
    {
      "KubeVersion": {
        "Version": "v1.20.0",
        "Major": "1",
        "Minor": "20"
      },
      "APIVersions": [
        "v1",
        "admissionregistration.k8s.io/v1",
        "admissionregistration.k8s.io/v1alpha1",
        "admissionregistration.k8s.io/v1beta1",
        "internal.apiserver.k8s.io/v1alpha1",
        "apps/v1",
        "apps/v1beta1",
        "apps/v1beta2",
        "authentication.k8s.io/v1",
        "authentication.k8s.io/v1alpha1",
        "authentication.k8s.io/v1beta1",
        "authorization.k8s.io/v1",
        "authorization.k8s.io/v1beta1",
        "autoscaling/v1",
        "autoscaling/v2",
        "autoscaling/v2beta1",
        "autoscaling/v2beta2",
        "batch/v1",
        "batch/v1beta1",
        "certificates.k8s.io/v1",
        "certificates.k8s.io/v1beta1",
        "certificates.k8s.io/v1alpha1",
        "coordination.k8s.io/v1beta1",
        "coordination.k8s.io/v1",
        "discovery.k8s.io/v1",
        "discovery.k8s.io/v1beta1",
        "events.k8s.io/v1",
        "events.k8s.io/v1beta1",
        "extensions/v1beta1",
        "flowcontrol.apiserver.k8s.io/v1",
        "flowcontrol.apiserver.k8s.io/v1beta1",
        "flowcontrol.apiserver.k8s.io/v1beta2",
        "flowcontrol.apiserver.k8s.io/v1beta3",
        "networking.k8s.io/v1",
        "networking.k8s.io/v1alpha1",
        "networking.k8s.io/v1beta1",
        "node.k8s.io/v1",
        "node.k8s.io/v1alpha1",
        "node.k8s.io/v1beta1",
        "policy/v1",
        "policy/v1beta1",
        "rbac.authorization.k8s.io/v1",
        "rbac.authorization.k8s.io/v1beta1",
        "rbac.authorization.k8s.io/v1alpha1",
        "resource.k8s.io/v1alpha2",
        "scheduling.k8s.io/v1alpha1",
        "scheduling.k8s.io/v1beta1",
        "scheduling.k8s.io/v1",
        "storage.k8s.io/v1beta1",
        "storage.k8s.io/v1",
        "storage.k8s.io/v1alpha1",
        "apiextensions.k8s.io/v1beta1",
        "apiextensions.k8s.io/v1"
      ],
      "HelmVersion": {
        "Version": "v3.14",
        "GitCommit": "",
        "GitTreeState": "",
        "GoVersion": "go1.24.10"
      }
    }
  chart: |-
    {
      "Version": "0.1.0",
      "Description": "",
      "Sources": [],
      "APIVersion": "v2",
      "Condition": "",
      "Type": "",
      "Name": "ts-test-chart",
      "AppVersion": "",
      "Keywords": [],
      "Home": "",
      "Icon": "",
      "Tags": "",
      "Annotations": {}
    }
  files-keys: |-
    [
      ".helmignore"
    ]
  release: |-
    {
      "IsUpgrade": false,
      "IsInstall": true,
      "Revision": 1,
      "Service": "Helm",
      "Name": "stub-release",
      "Namespace": "stub-namespace"
    }
  runtime: |-
    {
      "a": 1
    }
  values: |-
    {
      "testValue": "hello"
    }
kind: ConfigMap
metadata:
  annotations: {}
  labels: {}
  name: render-context-debug
```

ilya-lesikov:
@greptileai @coderabbitai

ts/nelm-types/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@nelm/types",
```
dmmordvi: @DmitryBochkarev @ilya-lesikov
npm package is ready - https://www.npmjs.com/package/@nelm/chart-ts-sdk
So we can remove `ts/nelm-types` content, it was moved into the package

PR #530, opened 2026-01-15
feat: implement local resource validation against api spec

drey:


cmd/nelm/chart_lint.go
```diff
@@ -74,6 +74,10 @@ func newChartLintCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*co
 			return fmt.Errorf("add chart repo connection flags: %w", err)
 		}
 
+		if err := AddLocalResourceValidationFlags(cmd, &cfg.ResourceLocalValidationOptions); err != nil {
```
ilya-lesikov: Please, grep the whole PR for "local" and get rid of it, like AddResourceValidationFlags

cmd/nelm/common_flags.go
```diff
@@ -295,6 +296,36 @@ func AddChartRepoConnectionFlags(cmd *cobra.Command, cfg *common.ChartRepoConnec
 	return nil
 }
 
+func AddLocalResourceValidationFlags(cmd *cobra.Command, cfg *common.ResourceLocalValidationOptions) error {
+	if !featgate.FeatGateLocalResourceValidation.Enabled() {
+		return nil
+	}
+
+	if err := cli.AddFlag(cmd, &cfg.NoResourceValidation, "no-validation", false, "Disable local resource validation", cli.AddFlagOptions{
```
ilya-lesikov: I just realised, we should name it `no-resource-validation` instead. I'm afraid of collisions down the line, there might be other types of validations

cmd/nelm/common_flags.go
```diff
@@ -295,6 +296,36 @@ func AddChartRepoConnectionFlags(cmd *cobra.Command, cfg *common.ChartRepoConnec
...
+		GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
+		Group:                resourceValidationGroup,
+	}); err != nil {
+		return fmt.Errorf("add flag: %w", err)
+	}
+
+	if err := cli.AddFlag(cmd, &cfg.ValidationKubeVersion, "validation-kube-version", common.DefaultLocalResourceValidationKubeVersion, "Kubernetes schemas to use during local resource validation", cli.AddFlagOptions{
```
ilya-lesikov: Same, "resource-validation-kube-version"

cmd/nelm/common_flags.go
```diff
@@ -295,6 +296,36 @@ func AddChartRepoConnectionFlags(cmd *cobra.Command, cfg *common.ChartRepoConnec
...
+		GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
+		Group:                resourceValidationGroup,
+	}); err != nil {
+		return fmt.Errorf("add flag: %w", err)
+	}
+
+	if err := cli.AddFlag(cmd, &cfg.ValidationSkip, "validation-skip", []string{}, "Skip local resource validation for resources with specified attributes", cli.AddFlagOptions{
```
ilya-lesikov: Same, "resource-validation-skip"

cmd/nelm/groups.go
```diff
@@ -21,4 +21,5 @@ var (
 	kubeConnectionFlagGroup = cli.NewFlagGroup("kube-connection", "Kubernetes connection options:", 50)
 	performanceFlagGroup    = cli.NewFlagGroup("performance", "Performance options:", 40)
 	miscFlagGroup           = cli.NewFlagGroup("misc", "Other options:", 0)
+	resourceValidationGroup = cli.NewFlagGroup("resource-local-validation", "Resource local validation options:", 75)
```
ilya-lesikov: Please, move it 6 lines higher (I'm sorry, I couldn't help it)

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+) error {
 	if err := validateNoDuplicates(releaseNamespace, transformedResources); err != nil {
 		return fmt.Errorf("validate for no duplicated resources: %w", err)
 	}
 
+	if !featgate.FeatGateLocalResourceValidation.Enabled() || opts.NoResourceValidation {
+		return nil
+	}
+
+	if err := validateResourceSchemas(ctx, releaseNamespace, transformedResources, opts); err != nil {
```
ilya-lesikov: Can you move it inside the `if featgate` block above, inverting the if logic? Otherwise some might not see this early return and add another validateSomething() in the end, and it will be toggled off with NoResourceValidation too.

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+		return fmt.Errorf("cannot initialize kubernetes %s cache: %w", signalFilePath, err)
+	}
+
+	defer f.Close()
+
 	return nil
 }
 
+func getKubeConformSchemaCacheDir(kubeVersion string, create bool) (string, error) {
+	cacheDirPath := helmpath.CachePath(".nelm", "api-resource-json-schemas", kubeVersion)
```
ilya-lesikov: The final path (by default) should be ~/.cache/helm/nelm/api-resource-json-schemas/...

ilya-lesikov: I'd also just do something like
```go
var APIResourceJSONSchemasCache = helmpath.CachePath(..., "api-resource-json-schemas")
```
somewhere in pkg/common/common.go and won't use getKubeConfromSchemaCacheDir anymore when I only need to get the path. The `create bool` argument can be remove too then, and the fuction renamed to createKubeConfromSchemaCacheDir

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+
+	if !create {
+		return cacheDirPath, nil
+	}
+
+	if stat, err := os.Stat(cacheDirPath); os.IsNotExist(err) {
+		if err := os.MkdirAll(cacheDirPath, 0o750); err != nil {
```
ilya-lesikov: We usually do 755 for dirs, 644 for files. Just checked, in some places we even do 777/666, which IS kinda stupid, but whatever. I'd say let's do 755 for consistency. The user shouldn't rely on this anyway, and do `chmod 750 /home/user` instead. There is also umask. Helm does 755/777 looks like.

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+}
+
+func checkIfKubeConformSpecExists(ctx context.Context, kubeVersion string) error {
+	cacheDir, err := getKubeConformSchemaCacheDir(kubeVersion, false)
+	if err != nil {
+		return err
+	}
+
+	// Use signal file to avoid sending subsequent requests
+	signalFilePath := filepath.Join(cacheDir, "ready")
```
ilya-lesikov: Hmm, I think we better do this HEAD request every time, just to be safe. This remote repo is not something we should rely, and even github is not something we should rely this much, so I don't think we need to cache this, so let's not overcomplicate. Maybe github will break something and will do 200 on every request even if there is nothing, or there are some misconfigured proxies on the user side, or someone will create something in the repo and then delete it, etc.

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+	} else if stat != nil {
+		return nil
+	}
+
+	// This file must exist to specific version. If it does not - version is not valid.
+	url := "https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v" + kubeVersion + "-standalone/all.json"
+
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
```
ilya-lesikov, line 140: Just saying, all of our kubeconform code in this PR should be able to work concurrently, e.g. multiple parallel action.ReleaseInstall() actions.

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("download schemas: %w", err)
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		return errors.New("schemas not found")
```
ilya-lesikov:
```suggestion
		return fmt.Errorf("schemas for kube version %q not found", kubeVersion)
```

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+	}
+
+	validatorOpts := validator.Opts{
+		Strict:               false, // Skip undefined params check
+		IgnoreMissingSchemas: true,
+		Cache:                cacheDir,
+		KubernetesVersion:    kubeVersion,
+		Debug:                true,
+	}
+
```
ilya-lesikov, line 121:
```suggestion

	validatorOpts := validator.Opts{
		Strict:               false, // Skip undefined params check
		IgnoreMissingSchemas: true,
		Cache:                cacheDir,
		KubernetesVersion:    kubeVersion,
	}

	if log.Default.AcceptLevel(ctx, log.DebugLevel) {
		validatorOpts.Debug = true
	}
```

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+INPUT:
+	for _, filter := range filters {
+		var match bool
+
+		properties, err := util.ParseProperties(ctx, filter)
+		if err != nil {
+			return false, fmt.Errorf("cannot parse properties: %w", err)
+		}
```
ilya-lesikov: We will validate these properties early (eventually :). Don't bother, there are already quite some panics in similar cases. We will hopefully add proper early validation for options in the next major
```suggestion

		properties := lo.Must(util.ParseProperties(ctx, filter))
```

cmd/nelm/common_flags.go
```diff
@@ -295,6 +296,36 @@ func AddChartRepoConnectionFlags(cmd *cobra.Command, cfg *common.ChartRepoConnec
...
+		GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
```
ilya-lesikov: Also, I guess we HAVE to explain the format in the help here, otherwise its impossible to figure it out. Something very-very short, like one-two short sentences.

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+
+		// check if all params are valid. Otherwise, operation can be unexpected, depending on iteration over properties.
+		for property := range properties {
+			if _, found := metaMap[property]; !found {
+				return false, fmt.Errorf("only %s skip properties are supported",
+					strings.Join(lo.Keys(metaMap), ", "))
+			}
+		}
```
ilya-lesikov: same here, don't bother with this validation for now

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+			f := metaMap[property]
+
+			if stringValue, ok := value.(string); !ok {
+				return false, fmt.Errorf("property %s has invalid value", property)
+			} else if !f(stringValue) {
+				continue INPUT
+			}
+
+			match = true
+		}
```
ilya-lesikov, line 203: Can't we reuse spec.ResourceMatcher for matching logic? Like here: https://github.com/werf/nelm/blob/82d465b9f46adbfc2e65987fb7b7093051cf2ba0/internal/resource/dependency.go#L335 

I intend to expand/improve ResourceMatcher over time, adding other kinds of filters (chart, label, regexp, etc)

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+	metaMap := map[string]func(string) bool{
+		"kind":    func(s string) bool { return s == meta.GroupVersionKind.Kind },
+		"group":   func(s string) bool { return s == meta.GroupVersionKind.Group },
+		"version": func(s string) bool { return s == meta.GroupVersionKind.Version },
+		"namespace": func(s string) bool {
+			if s == "" {
+				s = "default"
+			}
+
+			return s == meta.Namespace || (meta.Namespace == "" && s == releaseNamespace)
```
ilya-lesikov: The whole namespace thing in Nelm is kinda complicated. Nelm often needs to work with resources in no-cluster mode, but without a cluster we cannot know if the resource is namespaced. Down the line this creates problems with things like matching namespaces. I'm not gonna dive into how I solved this now, but let's just do this here:

```suggestion
			return s == meta.Namespace
```

internal/resource/validate.go
```diff
@@ -43,3 +270,65 @@ func validateNoDuplicates(releaseNamespace string, transformedResources []*Insta
...
+
+		if validationResult.Status == validator.Skipped {
+			log.Default.Debug(ctx, "Skip local validation for resource: %s", res.IDHuman())
+		}
+
+		if validationResult.Status != validator.Valid {
+			for _, validationError := range validationResult.ValidationErrors {
+				validationErrs = append(validationErrs, validationError.Path+": "+validationError.Msg)
+			}
+		}
```
ilya-lesikov: Hey, how about doing this via switch? This is mostly for safety and code discovery: in cases explicitly mention all valid values and panic on `default` (if any extra constant will be added now its gonna panic). As a bonus it's gonna be easier to understand this way

internal/resource/validate.go
```diff
@@ -43,3 +270,65 @@ func validateNoDuplicates(releaseNamespace string, transformedResources []*Insta
...
+
+	// Check for stream reading errors
+	for err := range errCh {
+		if err != nil {
+			return fmt.Errorf("read resource stream: %w", err)
+		}
+	}
+
+	if len(validationErrs) > 0 {
+		return fmt.Errorf("%s", strings.Join(validationErrs, "\n"))
```
ilya-lesikov: can you do multiline errors via util.Multierrorf helper?

internal/resource/validate.go
```diff
@@ -43,3 +270,65 @@ func validateNoDuplicates(releaseNamespace string, transformedResources []*Insta
...
+	return nil
+}
+
+func validateResourceWithCodec(res *InstallableResource) error {
+	unstruct := res.Unstruct
+	gvk := unstruct.GroupVersionKind()
+
+	jsonBytes, err := unstruct.MarshalJSON()
+	if err != nil {
+		return fmt.Errorf("marshal josn: %w", err)
```
ilya-lesikov:
```suggestion
		return fmt.Errorf("marshal json: %w", err)
```

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+	for _, res := range resources {
+		if ok, err := shouldSkipValidation(ctx, opts.ValidationSkip, releaseNamespace, res.ResourceMeta); err != nil {
+			return fmt.Errorf("skip validation: %w", err)
+		} else if ok {
+			log.Default.Debug(ctx, "Skip local validation for resource (due to ValidationSkip): %s", res.IDHuman())
+
+			continue
+		}
+
+		if err := validateResourceSchemaWithKubeConform(ctx, resValidator, res); err != nil {
```
ilya-lesikov, line 79: This function must somehow separate regular fatal errors and validation errors. There are few ways to do that.

How about creating struct KubeConformValidationError (or just var ... = errors.New()) which implements interface error and return it from inside the function only when its really a validation error.

For all other errors return plain error.

If we got plain error: that's fatal, just stop the execution and abort. If we got validation error: collect them, then format and print them in the end.

internal/resource/validate.go
```diff
@@ -1,23 +1,250 @@
...
+			for _, s := range strings.Split(err.Error(), "\n") {
+				sb.WriteString("  validate " + res.IDHuman() + ": " + s + "\n")
+			}
+
+			continue
+		}
+
+		if err := validateResourceWithCodec(res); err != nil &&
+			!strings.Contains(err.Error(), fmt.Sprintf("no kind %q is registered", res.GroupVersionKind.Kind)) {
```
ilya-lesikov: Same, probably CodecValidationError needed in validateResourceWithCodec. Figuring out what kind of error is this (fatal or regular "no kind") must happen inside of validateResourceWithCodec.

internal/resource/validate.go
drey: Refactored as discussed on the call.

internal/resource/validate.go
drey: I am afraid util.Multierrorf will not allow to print errors in the target format, as this function concat all the errors in the single line.

internal/resource/validate.go
drey: I agree on above recommendation regarding KubeConform, but in this case it is hard to deviate fatal and regular errors, as an error like "no kind" is not fatal by its nature, it just the specific custom resource case. So, in my opinion it is better to suppress errors like "no kind" (return nil) and do not introduce additional wrapping.

internal/resource/validate.go
ilya-lesikov: I think it should:
```
Error: install: remotely validate resources: validate adoptable resources: adoption validation failed: 2 errors occurred:
	* resource "ConfigMap/myccmmm" is not adoptable: annotation "meta.helm.sh/release-name=myreleasea" must have value "myrelease"
	* resource "ConfigMap/myccmmm2" is not adoptable: annotation "meta.helm.sh/release-namespace=myreleasea" must have value "myrelease"
```

internal/resource/validate.go
ilya-lesikov: I was thinking doing it like this (pseudocode):
```
func validateResourceWithCodec() error {
  if something fatal return err
  if "no kind" return nil
  if other Decode() error return ValidationError
}
```
Or am I missing something?

drey: What do you mean by something fatal?

internal/resource/validate.go
```diff
@@ -17,34 +17,33 @@ import (
...
 
-	"github.com/werf/3p-helm/pkg/helmpath"
 	"github.com/werf/nelm/internal/resource/spec"
 	"github.com/werf/nelm/internal/util"
 	"github.com/werf/nelm/pkg/common"
 	"github.com/werf/nelm/pkg/featgate"
 	"github.com/werf/nelm/pkg/log"
 )
 
+var ErrKubeConformUnexpectedError = errors.New("unexpected non-validation error")
```
ilya-lesikov: Can we invert it, I mean to have ErrKubeConformValidation instead, like proposed initially? I think the general rule should be "no special error if nothing special happening, i.e. the error is fatal and immediate".

internal/resource/validate.go
```diff
@@ -17,34 +17,33 @@ import (
...
+
 // Can be called even without cluster access.
 func ValidateLocal(ctx context.Context, releaseNamespace string, transformedResources []*InstallableResource,
-	opts common.ResourceLocalValidationOptions,
+	opts common.ResourceValidationOptions,
```
ilya-lesikov, line 43: That's probably some formatting in IDE or idk, or is this the wormatter? Opts shouldn't hang like that

drey: I have no idea why it looks so. Can you refresh the page? I already see `opts common.ResourceValidationOptions` in my code as well as in the changed files tab of the PR.

internal/resource/validate.go
```diff
@@ -153,91 +139,60 @@ func checkIfKubeConformSpecExists(ctx context.Context, kubeVersion string) error
...
-			}
-
-			return s == meta.Namespace || (meta.Namespace == "" && s == releaseNamespace)
-		},
-		"name": func(s string) bool { return s == meta.Name },
-	}
-
-INPUT:
+func shouldSkipValidation(ctx context.Context, filters []string, releaseNamespace string,
+	meta *spec.ResourceMeta,
```
ilya-lesikov: strange formatting again

drey: I got your point. I do it to minimise line length. Usually it is a general recommendation. You want me not to do so?

internal/resource/validate.go
```diff
@@ -286,25 +241,27 @@ func validateResourceSchemaWithKubeConform(ctx context.Context, kubeConformValid
...
+			log.Default.Debug(ctx, "Skip validation for resource: %s", res.IDHuman())
+		case validator.Invalid:
 			for _, validationError := range validationResult.ValidationErrors {
 				validationErrs = append(validationErrs, validationError.Path+": "+validationError.Msg)
 			}
+		case validator.Empty, validator.Valid:
+			continue
+		default:
+			return fmt.Errorf("%w: %w", ErrKubeConformUnexpectedError,
+				fmt.Errorf("unexpected validation status: %d", validationResult.Status))
```
ilya-lesikov: Just the stupidest panic would suffice, and don't even bother with sprintf (or do, I'm too lazy personally)
```suggestion
				panic("unexpected validation status")
```

pkg/common/common.go
```diff
@@ -296,3 +296,11 @@ var (
 )
 
 var SprigFuncs = sprig.TxtFuncMap()
+
+var (
+	APIResourceValidationJSONSchemasCacheDir        = helmpath.CachePath("nelm", "api-resource-json-schemas")
+	APIResourceValidationKubeConformSchemasLocation = []string{
+		"default",
+		"https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json",
```
ilya-lesikov, line 304: It's all fine, but in the future probably better to move the second link to a separate var, so when you need to pick it from this list and modify it or something, you don't need to pick it by index, which is very brittle. The "default" is probably doesn't even need a constant (or maybe it has constant somewhere inside 3rd-party lib)

internal/resource/validate.go
ilya-lesikov: I guess you are right that it's not _required_ right now, but will complicate updating this function. By fatal I meant  something like [this](https://github.com/werf/nelm/blob/89e379f08cb2a5ecd1d4d0104225a414e22ba20d/internal/resource/validate.go#L280) (at this point something is very wrong if it can't marshal, might even be a programmer error?) or maybe some non-validation errors will surface in [this](https://github.com/werf/nelm/blob/89e379f08cb2a5ecd1d4d0104225a414e22ba20d/internal/resource/validate.go#L287) soon. Other operations might be added later that will return non-validation errors, would be good if it won't require a refactor for this and you could just return a regular "error" as we always do.

internal/resource/validate.go
ilya-lesikov: Idk, I refreshed the page and it still shows like a separate line with opts. In the diff tab I do see a plus to the left of it, which I guess means a new line.

internal/resource/validate.go
ilya-lesikov: Oh, alright. I decided to follow this: all args always on a single line (whatever the length is). There is wrapping in any ide/editor anyways. But if the line is too long, it probably needs a refactor by grouping some things in structs or something.

I guess the other way could be to just one arg per line, which is very formatting and git-friendly, but it really bloats files, making it hard to follow. And if there are too much args we should refactor anyway.

PR #533, opened 2026-01-16
fix: new patcher for Secret, which coverts stringData to data

dmmordvi:


internal/resource/spec/patch.go
```diff
@@ -132,3 +135,31 @@ func (p *LegacyOnlyTrackJobsPatcher) Patch(ctx context.Context, info *ResourcePa
...
+func NewSecretStringDataPatcher() *SecretStringDataPatcher {
+	return &SecretStringDataPatcher{}
+}
+
+func (p *SecretStringDataPatcher) Match(ctx context.Context, info *ResourcePatcherResourceInfo) (bool, error) {
+	return info.Obj.GetKind() == "Secret", nil
+}
+
+func (p *SecretStringDataPatcher) Patch(ctx context.Context, info *ResourcePatcherResourceInfo) (*unstructured.Unstructured, error) {
+	unstructObject := info.Obj.UnstructuredContent()
```
ilya-lesikov: Whenever possible, prefer working with Unstructured{} directly, without converting it to map[string]interface or json. There is a bunch of helpers for this in the [unstructured](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured) package (getters, setters).

Working with arbitrary yaml/json in Go is a pain, so that's why k8s devs made unstructured package.

internal/resource/spec/patch.go
```diff
@@ -132,3 +135,31 @@ func (p *LegacyOnlyTrackJobsPatcher) Patch(ctx context.Context, info *ResourcePa
...
 }
+
+type SecretStringDataPatcher struct{}
+
+func NewSecretStringDataPatcher() *SecretStringDataPatcher {
+	return &SecretStringDataPatcher{}
+}
+
+func (p *SecretStringDataPatcher) Match(ctx context.Context, info *ResourcePatcherResourceInfo) (bool, error) {
+	return info.Obj.GetKind() == "Secret", nil
```
ilya-lesikov: Also check for Group, it should be exactly empty string (`""`). This way we will match the builtin Secret resource, and never match any other custom resource from 3rd-party providers.

internal/resource/spec/patch.go
```diff
@@ -132,3 +135,31 @@ func (p *LegacyOnlyTrackJobsPatcher) Patch(ctx context.Context, info *ResourcePa
...
+}
+
+func (p *SecretStringDataPatcher) Patch(ctx context.Context, info *ResourcePatcherResourceInfo) (*unstructured.Unstructured, error) {
+	unstructObject := info.Obj.UnstructuredContent()
+
+	if stringData, ok := unstructObject["stringData"].(map[string]interface{}); ok {
+		data := unstructObject["data"].(map[string]interface{})
```
ilya-lesikov: I'm not sure, but I think there might be a panic if there is no "data" field or if it is not map[string]interface, but string or something else (which is user error, thus we shouldn't panic).

ilya-lesikov: If there is no data we should make a new "data" (`data := map[string]interface{}{}`)

pkg/action/release_install.go
```diff
@@ -355,6 +355,7 @@ func releaseInstall(ctx context.Context, ctxCancelFn context.CancelCauseFunc, re
 
 	patchers := []spec.ResourcePatcher{
 		spec.NewExtraMetadataPatcher(opts.ExtraAnnotations, opts.ExtraLabels),
+		spec.NewSecretStringDataPatcher(),
```
ilya-lesikov, line 358: You know, let's add it also to BuildReleasableResourceSpecs() in runRollbackPlan().

internal/resource/spec/patch.go
ilya-lesikov: Actually, you can even use [NestedStringMap](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#NestedStringMap) helper. We don't need map[string]interface{} here, these fields should always be map[string]string. Unless there is a user error (in which case we should do nothing and continue)

internal/resource/spec/patch.go
```diff
@@ -132,3 +137,47 @@ func (p *LegacyOnlyTrackJobsPatcher) Patch(ctx context.Context, info *ResourcePa
...
+
+func (p *SecretStringDataPatcher) Match(ctx context.Context, info *ResourcePatcherResourceInfo) (bool, error) {
+	return info.Obj.GroupVersionKind().GroupKind() == schema.GroupKind{Group: "", Kind: "Secret"}, nil
+}
+
+func (p *SecretStringDataPatcher) Patch(ctx context.Context, info *ResourcePatcherResourceInfo) (*unstructured.Unstructured, error) {
+	content := info.Obj.UnstructuredContent()
+
+	data, found, err := unstructured.NestedStringMap(content, "data")
+	if err != nil {
```
ilya-lesikov, line 155: I think if `.data` is not a map[string]string, then you'll get an error here, not just empty data. Probably you should check for error here and if err then return nil, nil. Let if fail later on general resource validation.

internal/resource/spec/patch.go
```diff
@@ -132,3 +137,47 @@ func (p *LegacyOnlyTrackJobsPatcher) Patch(ctx context.Context, info *ResourcePa
...
+		return nil, fmt.Errorf("read .data: %w", err)
+	}
+
+	if !found || data == nil {
+		data = map[string]string{}
+	}
+
+	stringData, found, err := unstructured.NestedStringMap(content, "stringData")
+	if err != nil {
+		return nil, fmt.Errorf("read .stringData: %w", err)
```
ilya-lesikov:
```suggestion
		return nil, nil
```

internal/resource/spec/patch.go
```diff
@@ -132,3 +137,47 @@ func (p *LegacyOnlyTrackJobsPatcher) Patch(ctx context.Context, info *ResourcePa
...
+	}
+
+	if found && stringData != nil {
+		for key, val := range stringData {
+			data[key] = base64.StdEncoding.EncodeToString([]byte(val))
```
ilya-lesikov, line 170: Check please what happens if there is a same key in both data and stringData. How Kubernetes handles this case by default, what takes priority? We need to preserve Kubernetes behavior.

dmmordvi: Checked, Kubernetes handles such case the same way, just rewrites existing `data` field using value from `stringData`. So it should work fine

PR #538, opened 2026-01-22
feat: webhook configuration deploy dependency

dmmordvi:


internal/resource/metadata.go
```diff
@@ -916,15 +916,25 @@ func deleteOnFailed(meta *spec.ResourceMeta) bool {
 	return lo.Contains(deletePolicies, common.DeletePolicyFailed)
 }
 
+func getWebhookStage(meta *spec.ResourceMeta) common.Stage {
+	if _, found := spec.FindAnnotationsOrLabelsByKeyPattern(meta.Annotations, common.AnnotationKeyPatternDeployDependency); found {
+		// If webhook has deploy-dependency annotation - use standard StageInstall, as for other resources
```
ilya-lesikov: Don't bother with comments, this code is straightforward

internal/resource/metadata.go
```diff
@@ -916,15 +916,25 @@ func deleteOnFailed(meta *spec.ResourceMeta) bool {
 	return lo.Contains(deletePolicies, common.DeletePolicyFailed)
 }
 
+func getWebhookStage(meta *spec.ResourceMeta) common.Stage {
+	if _, found := spec.FindAnnotationsOrLabelsByKeyPattern(meta.Annotations, common.AnnotationKeyPatternDeployDependency); found {
```
ilya-lesikov: Let's not touch the annotation directly here. It's a job of manualInternalDependencies() to figure it out. Just pass the result of manualInternalDependencies() down to deployConditions, like we do in weight() function here.

PR #539, opened 2026-01-22
feat: annotation `werf.io/delete-dependency-<id>`

ilya-lesikov:


internal/resource/dependency.go
```diff
@@ -102,6 +102,8 @@ func internalDependencies(unstruct *unstructured.Unstructured) []*InternalDepend
 	return dependencies
 }
 
+func externalDeployDependences() {}
```
ilya-lesikov: That's not what I meant. There should be internalDeleteDependencies()

internal/plan/plan_build.go
```diff
@@ -48,6 +48,10 @@ func BuildPlan(installableInfos []*InstallableResourceInfo, deletableInfos []*De
 		return plan, fmt.Errorf("connect internal dependencies: %w", err)
```
ilya-lesikov, line 48: lets rename connectInternalDependencies to connectInternalDeployDependencies

internal/resource/resource.go
```diff
@@ -190,12 +192,20 @@ func NewDeletableResource(spec *spec.ResourceSpec, releaseNamespace string, opts
 		delPropagation = deletePropagation(spec.ResourceMeta, opts.DefaultDeletePropagation)
 	}
 
-	return &DeletableResource{
-		ResourceMeta:      spec.ResourceMeta,
-		Ownership:         owner,
-		KeepOnDelete:      keep,
-		DeletePropagation: delPropagation,
+	if err := validateDeleteDependencies(spec.ResourceMeta); err != nil {
```
ilya-lesikov: Since a deletable resource sometimes is out of our control, i.e. the only thing that we can do with it is to delete it, then all errors in this constructor, if made fatal, can brick Helm releases. That's why we avoid making them fatal.

If you look above, if we got an error from validations in this constructor, we are falling back to some default and ignoring an error. Let's do the same here

PR #540, opened 2026-01-22
refactor: new MultiError

ilya-lesikov:


internal/util/multierror.go
```diff
@@ -6,14 +6,57 @@ import (
...
+				l.err = multierror.Append(l.err, other.err.Errors...)
+			}
+		} else {
+			l.err = multierror.Append(l.err, err)
+		}
+	}
+}
+
+func (l *MultiError) Error() string {
+	if l == nil || l.err == nil {
```
drey: Seems `l == nil` is not needed

PR #545, opened 2026-01-26
feat: enhanced local resource validation

drey:
<div id='description'>
<h3>Summary by Bito</h3>

<p>This pull request enhances the testing infrastructure for local resource validation by refactoring test helpers and adding comprehensive AI-generated test suites for kubeconform and general validation.</p>

<details>
<ul>

<li>Refactored test constants in helpers_ai_test.go to use a const block for better code organization.</li>

<li>Introduced new utility functions newSchemaHandler and setupDefaultSchemaServer in helpers_ai_test.go to reduce duplication in test setups.</li>

<li>Added new AI test file kubeconform_ai_test.go with multiple test cases covering valid resources, invalid resources, edge cases, and integration scenarios for kubeconform validation.</li>

<li>Introduced AI test suite in validate_ai_test.go for resource validation including codec validation, duplicate detection, namespace protection, and validation skipping.</li>

</ul>
</details>

</div>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,6 +1,6 @@
 # Repository Guidelines
 
-Nelm is a Go-based Kubernetes deployment tool (Helm-compatible). For context and expectations, read `README.md`, `CONTRIBUTING.md`, and `docs/summary/index.md`.
+Nelm is a Go-based Kubernetes deployment tool, which deploys Helm charts, is compatible with Helm releases, and is an alternative to Helm. For context and expectations, read `README.md`, `CONTRIBUTING.md`, and `docs/summary/index.md`.
 
 ## Documentation
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,10 +18,6 @@ For small changes, such as few lines bugfixes or documentation improvements, fee
 
 For easy first issues, check the [good first issue](https://github.com/werf/nelm/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22) tag.
 
-## Architecture
-
-See [ARCHITECTURE.md](ARCHITECTURE.md).
-
 ## Development
 
 1. Clone the repository.
@@ -102,7 +98,7 @@ Errors:
 * Always wrap errors with additional context using `fmt.Errorf("...: %w", err)`.
 * On programmer errors prefer panics, e.g. on an unexpected case in a switch.
 * Do one-line `if err := myfunc(); err != nil` wherever possible.
-* When wrapping errors with fmt.Errorf, describe what is being done, not what failed, e.g. `fmt.Errorf("reading config file: %w", err)` instead of `fmt.Errorf("cannot read config file: %w", err)`.
+* When wrapping errors with fmt.Errorf, describe what is being done, not what failed, e.g. `fmt.Errorf("read config file: %w", err)` instead of `fmt.Errorf("cannot read config file: %w", err)`.
 
 Concurrency:
 * Instead of raw mutexes use the transactional `Concurrent{}` helper.


cmd/nelm/common_flags.go
```diff
@@ -323,6 +330,29 @@ func AddResourceValidationFlags(cmd *cobra.Command, cfg *common.ResourceValidati
 		return fmt.Errorf("add flag: %w", err)
 	}
 
+	if err := cli.AddFlag(cmd, &cfg.ValidationSchema, "resource-validation-schema", common.DefaultResourceValidationSchema, "Default json schema sources to validate resources. Must be a valid go template defining a http(s) URL, or an absolute path on local file system", cli.AddFlagOptions{
+		GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
```
ilya-lesikov:
```suggestion
		GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalMultiEnvVarRegexes,
```

cmd/nelm/common_flags.go
```diff
@@ -323,6 +330,29 @@ func AddResourceValidationFlags(cmd *cobra.Command, cfg *common.ResourceValidati
...
+		Group:                resourceValidationGroup,
+		NoSplitOnCommas:      true,
+	}); err != nil {
+		return fmt.Errorf("add flag: %w", err)
+	}
+
+	if err := cli.AddFlag(cmd, &cfg.ValidationExtraSchema, "resource-validation-extra-schema", []string{}, "Extra json schema sources to validate resources (preferred over default sources). Must be a valid go template defining a http(s) URL, or an absolute path on local file system", cli.AddFlagOptions{
+		GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
```
ilya-lesikov:
```suggestion
		GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalMultiEnvVarRegexes,
```

cmd/nelm/common_flags.go
```diff
@@ -323,6 +330,29 @@ func AddResourceValidationFlags(cmd *cobra.Command, cfg *common.ResourceValidati
...
+		Group:                resourceValidationGroup,
+		NoSplitOnCommas:      true,
+	}); err != nil {
+		return fmt.Errorf("add flag: %w", err)
+	}
+
+	if err := cli.AddFlag(cmd, &cfg.ValidationSchemaCacheLifetime, "resource-validation-cache-lifetime", common.DefaultResourceValidationCacheLifetime, "How long local schema cache will be valid", cli.AddFlagOptions{
```
ilya-lesikov, line 349: If their downloading takes like 2 second, then ok. But otherwise maybe increase lifetime to up to a week? See for yourself

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,665 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("create schema validator: %w", err)
+	}
+
+	return &kubeConformLocalFSValidatorInstance{
+		validator: validatorInstance,
+	}, nil
+}
+
+func (v kubeConformLocalFSValidatorInstance) AddEntry(_ context.Context, _ schema.GroupVersionKind) error {
```
ilya-lesikov: I'd say let's use pointer methods by default, unless needed otherwise. Frankly, I can't remember if I ever needed non-pointer method in the whole Nelm codebase...
```suggestion
func (v *kubeConformLocalFSValidatorInstance) AddEntry(_ context.Context, _ schema.GroupVersionKind) error {
```

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,665 @@
...
+	lockFilePath := filepath.Join(cacheDir, "lock")
+
+	v := &kubeConformHTTPValidatorInstance{
+		cacheDir:      cacheDir,
+		cacheLifetime: cacheLifetime,
+		fileLock:      flock.New(lockFilePath),
+		validator:     validatorInstance,
+	}
+
+	err = v.fileLock.Lock()
```
ilya-lesikov: Please, squash all of these `if err :=` blocks ... or redo the whole Nelm repo so we never use the short form :)

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,665 @@
...
+	KindSuffix                  string
+	NormalizedKubernetesVersion string
+	ResourceAPIVersion          string
+	ResourceKind                string
+	StrictSuffix                string
+}
+
+type kubeConformCacheMetadata struct {
+	Entries map[string]time.Time `json:"entries"`
+	Version int                  `json:"version"`
```
ilya-lesikov: I already have a few of these in different places, so let's keep them consistent with each other (value here should be "v1"):
```suggestion
	APIVersion string                  `json:"apiVersion"`
```

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,665 @@
...
+			return nil, fmt.Errorf("unable to open metadata file: %w", err)
+		}
+
+		defer metadataFile.Close()
+
+		meta := kubeConformCacheMetadata{
+			Entries: make(map[string]time.Time),
+		}
+
+		encoder := json.NewEncoder(metadataFile)
```
ilya-lesikov: I thought you gonna decode/encode yaml here. Not much difference, but since we call it .yaml maybe let's encode/decode to the usual yaml instead of pure json

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,665 @@
...
+}
+
+type kubeConformValidatorRegistry struct {
+	mutex         sync.RWMutex
+	validatorsMap map[string]struct{}
+	validators    []kubeConformValidatorInstanceInterface
+}
+
+func (r *kubeConformValidatorRegistry) FindEntry(ctx context.Context, gvk schema.GroupVersionKind) (kubeConformValidatorInstanceInterface, bool, error) {
+	r.mutex.RLock()
```
ilya-lesikov: Let's use util.NewConcurrent() instead of manual mutex manipulation. I used util.Concurrent{} extensively in kubedog, and there was a lot of concurrency, and it worked great. Probably the only con is that there is a bit of a boilerplate on the client side when using .Transaction(), but it's worth it.

internal/resource/kubeconform.go
drey: Good catch. Changed file name to metadata.json.

internal/resource/kubeconform.go
drey: Went through util.NewConcurrent() it has a good fit where we want to protect some custom types' attribute access/change, but seems like less practical in the cases where custom types methods are involved. Let's discuss design decisions on a meeting.

drey: Decided not to use a dedicated Registry construct at all, so mutex is not needed anymore.

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+	"github.com/werf/nelm/pkg/log"
+)
+
+const (
+	kubeConformCacheLockFilename       = "lock"
+	kubeConformCacheMetadataAPIVersion = "1"
+	kubeConformCacheMetadataFilename   = "metadata.json"
+)
+
+type KubeConformValidator struct {
```
ilya-lesikov: Let's make it private. We don't need it now outside of the package. Constructor too

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+
+func NewKubeConformValidator(kubeVersion string, schemaCacheLifetime time.Duration, schemaSource []string) (*KubeConformValidator, error) {
+	return &KubeConformValidator{
+		kubeVersion:         strings.TrimLeft(kubeVersion, "v"),
+		schemaCacheLifetime: schemaCacheLifetime,
+		schemaSources:       schemaSource,
+	}, nil
+}
+
+func (kc *KubeConformValidator) Validate(ctx context.Context, res *InstallableResource) error {
```
ilya-lesikov: I don't think we'll ever need anything more than ResourceSpec in here, so let's get rid of InstallableResource in favor of ResourceSpec. The general rule would be to pass as simple/limited stuff as possible to functions. Less entropy, clearer, simpler

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+	}
+
+	return kc.validators, nil
+}
+
+// validateSchemasSources validates source against Kubernetes version based on assumption that
+// every Kubernetes version has Deployment.apps/v1 schema. If at least one source statisfy to condition, the complete
+// source list considered to be valid, as it allows to validate native resources.
+func (kc *KubeConformValidator) validateSchemasSources(ctx context.Context) error {
+	schemaSourceParams := kubeConformSchemaSourceParams{
```
ilya-lesikov: if a few lines below really is the only place it's used, I think having a whole struct for this isn't worth it. I think we should just inline all of these params as args to patchKubeConformSchemaSource.

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+		ResourceKind:                "deployment",
+		StrictSuffix:                "",
+	}
+
+	errs := &util.MultiError{}
+
+	for _, schemaSource := range kc.schemaSources {
+		patchedSource, err := patchKubeConformSchemaSource(schemaSource, schemaSourceParams)
+		if err != nil {
+			errs.Add(fmt.Errorf("patch schema source with test params: %w", err))
```
ilya-lesikov: I'd say we should only use MultiError when there can be a lot of errors. Like when we lint a 100 resources and each might fail in a different way. Then without multierror the user will really struggle fixing all of this.

Here though it's hard to imagine having more than 5 sources for 99.9% of people. For 98% people it'll be just the default sources. So no need to overcomplicate with multierror imo.

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+		return kc.validators, nil
+	}
+
+	if err := kc.validateSchemasSources(ctx); err != nil {
+		return nil, fmt.Errorf("validate schema sources: %w", err)
+	}
+
+	// Generate top level directory name based on source combination, to avoid invalid cache hit on
+	// source combination change.
+	sourcesSubDirName := getHash(strings.Join(kc.schemaSources, "-"))
```
ilya-lesikov, line 155: There is a more proper way to compute sha256 in this case by sha256.New() + hash.Write(). Doesn't really matter though

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+	"sigs.k8s.io/yaml"
+
+	"github.com/werf/nelm/internal/util"
+	"github.com/werf/nelm/pkg/common"
+	"github.com/werf/nelm/pkg/log"
+)
+
+const (
+	kubeConformCacheLockFilename       = "lock"
+	kubeConformCacheMetadataAPIVersion = "1"
```
ilya-lesikov: "v1" please

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+		if err := encoder.Encode(meta); err != nil {
+			return nil, fmt.Errorf("encode %s: %w", metadataFilePath, err)
+		}
+
+		v.metadata = meta
+
+		return v, nil
+	}
+
+	metadataFile, err := os.OpenFile(metadataFilePath, os.O_RDONLY, 0o644)
```
ilya-lesikov: You can probably deduplicate by using loadMetadataFromDisk() here. Also, I'd be making func loadMetadataFromDisk a function, not a method

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+	KindSuffix                  string
+	NormalizedKubernetesVersion string
+	ResourceAPIVersion          string
+	ResourceKind                string
+	StrictSuffix                string
+}
+
+type kubeConformCacheMetadata struct {
+	APIVersion string               `json:"apiVersion"`
+	Entries    map[string]time.Time `json:"entries"`
```
ilya-lesikov: Hey, maybe change the structure a bit:
```yaml
apiVersion: xxx
entries:
- hash: xxx
  timestamp: xxx
```
Otherwise it'll be a pain in the ass to extend.

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+	}()
+
+	if err := v.loadMetadataFromDisk(); err != nil {
+		return fmt.Errorf("refresh metadata from %s: %w", v.metadataFilePath(), err)
+	}
+
+	var changed bool
+
+	for hash, timestamp := range v.metadata.Entries {
+		if timestamp.Add(v.cacheLifetime).Before(time.Now().UTC()) {
```
ilya-lesikov: early return would be nice

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+	ResourceKind                string
+	StrictSuffix                string
+}
+
+type kubeConformCacheMetadata struct {
+	APIVersion string               `json:"apiVersion"`
+	Entries    map[string]time.Time `json:"entries"`
+}
+
+type kubeConformValidatorInstance struct {
```
ilya-lesikov: Yea, naming is tough here... Can't even come up with anything. Maybe just kubeConformInstance?

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+	} else if err != nil {
+		return "", fmt.Errorf("stat cache dir %q: %w", path, err)
+	} else if !stat.IsDir() {
+		return "", fmt.Errorf("%s is not a directory", path)
+	}
+
+	return path, nil
+}
+
+func getKubeConformEntryHash(kubeVersion string, gvk schema.GroupVersionKind) string {
```
ilya-lesikov, line 473: This is the hash of the actual json file? I guess we don't need kubeVersion here?

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+	if err != nil {
+		return fmt.Errorf("marshal resource to yaml: %w", err)
+	}
+
+	validators, err := kc.getValidatorInstances(ctx)
+	if err != nil {
+		return fmt.Errorf("get validators: %w", err)
+	}
+
+	var cachedEntryFound bool
```
ilya-lesikov: I guess you can reuse "found" below, just rename it

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+	if matchedValidator, found, err := kc.findCachedEntry(ctx, res.GroupVersionKind); err != nil {
+		return fmt.Errorf("get validator: %w", err)
+	} else if found {
+		cachedEntryFound = true
+		validators = []*kubeConformValidatorInstance{matchedValidator}
+	}
+
+VALIDATOR:
```
ilya-lesikov: just for consistency:

```suggestion
validatorLoop
```

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+	for _, schemaValidator := range validators {
+		validationErrs := &util.MultiError{}
+
+		resCh, errCh := resource.FromStream(ctx, res.FilePath, bytes.NewReader(yamlBytes))
```
ilya-lesikov: Please, add the comment above this line, so we won't forget in v2:

```
// TODO(v2): if possible, we should use only a single yaml marshaller and a single json marshaller everywhere
```

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+	defer func() {
+		if err := v.fileLock.Unlock(); err != nil {
+			log.Default.Error(ctx, "release lock on schema validator %s: %s", v.lockFilePath(), err)
+		}
+	}()
+
+	// Underlying calls has thread unsafe os.WriteFile call, so we must protect it by file lock.
+	result := v.validator.ValidateResource(res)
+
+	return &result, nil
```
ilya-lesikov: There is a nice helper for this: lo.ToPtr()

lo lib in general has a few nice helpers, especially for working with maps/slices

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+}
+
+func (v *kubeConformValidatorInstance) dumpMetadata() error {
```
ilya-lesikov: I don't quite get what this function does

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+					validationErrs.Add(fmt.Errorf("%s: %w", validationErr.Path, &validationErr))
+				}
+			case validator.Empty, validator.Valid:
+				if !cachedEntryFound {
+					if err := schemaValidator.AddCacheEntry(ctx, res.GroupVersionKind); err != nil {
+						return fmt.Errorf("add entry %s: %w", res.IDHuman(), err)
+					}
+				}
+
+				continue
```
ilya-lesikov: redundant

internal/resource/kubeconform.go
```diff
@@ -0,0 +1,565 @@
...
+				if !cachedEntryFound {
+					if err := schemaValidator.AddCacheEntry(ctx, res.GroupVersionKind); err != nil {
+						return fmt.Errorf("add entry %s: %w", res.IDHuman(), err)
+					}
+				}
+
+				for _, validationErr := range validationResult.ValidationErrors {
+					validationErrs.Add(fmt.Errorf("%s: %w", validationErr.Path, &validationErr))
+				}
+			case validator.Empty, validator.Valid:
```
ilya-lesikov: `if !cachedEntryFound` is never triggered there for Empty? Or this is a bug

Anyways, I'd say let's panic on Empty: it should never be empty that far into the execution for InstallableResource. Something should be very broken then

ilya-lesikov:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

internal/resource/kubeconform.go
drey: This method writes metadata to disk. Changed function name to writeMetadata().

internal/resource/kubeconform.go
drey: This code involved only once until `kc.validators` has no validators. It won't be involved for upcoming resource validations.

During initialisation we need to check that at least one of the sources has a json spec for deployment.apps/v1 to be sure that specific source has native schemas for specified kube version. If no sources meet the condition we have multiple errors. In my opinion it makes sense to print error for every source with exact URL that was used during the check. This is why I use multi error here. See an example below (unsupported kube-version specified):

```
 NELM_FEAT_RESOURCE_VALIDATION=true nelm release plan install --namespace local-resource-validation --release local-resource-validation .helm --log-level=debug --resource-validation-kube-version 1.3211323.0
Planning release install "local-resource-validation" (namespace: "local-resource-validation")
Build release history
Render chart
Merging override values for chart at ".helm"
Loading chart at ".helm"
Rendering values for chart at ".helm"
Rendering resources for chart at ".helm"
Build transformed resource specs
Build releasable resource specs
Convert previous release to resource specs
Convert new release to resource specs
Build resources
Locally validate resources
Error: release plan install: locally validate resources: validate resource schemas: validate Deployment/nginx-deployment1: get validators: validate schema sources: 2 errors occurred:
  * test schema https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v1.3211323.0-standalone/deployment.json for kube version 1.3211323.0 not found
  * test schema https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/apps/deployment_v1.json for kube version 1.3211323.0 not found
```

The same sort of errors will occur if both --resource-validation-scheme and --resource-validation-extra-scheme provided by user do not meet the above condition.

ilya-lesikov: Frankly I didn't get the first part of your response, how it is related to multierror.

The point I made is if we use MultiError in any place that can theoretically loop and produce more than 1 error, it will be a lot of places, and a lot of boilerplate.

The goal of multierror is not to show as much errors as possible in any situation, but make it easier for the user to fix his stuff when he got a lot of errors, especially different errors, all at once. So that he doesn't restart nelm many times in the row, fixing each error for every incorrect resource in the release, for example.

This problem exists when there are like dozens of errors at once. When there are just 2-3 errors (and especially, like in your example, when they are effectively the same error), the issue of "that's gonna be a struggle to fix 100 errors one-by-one" is non-existent. That's why I think code simplicity should win over here.

internal/resource/kubeconform.go
drey: Using list as a construct leads to O(N) time complexity. Using hash map is near O(1).

internal/resource/kubeconform.go
drey: Sorry, I didn't get your point. Here we want t o invalidate all expired entries, don't see how early return can be used here.

internal/resource/kubeconform.go
drey: We do need it, because we have the following directory tree:

```
<Hash of all sources regardless of kube version>
  <hash of exact source regardless of kube version>
     <hash exact schema with kube version>
```

internal/resource/kubeconform.go
drey: The current metadata looks so:

```
{
  "apiVersion": "v1",
  "entries": {
    "1a63bcc44687cac3f168f397affdac8ecedb5237e42460908a0827950d1a3288": {
      "created": "2026-01-30T16:07:34.072517Z"
    }
  }
}
```

internal/resource/kubeconform.go
ilya-lesikov: I meant early continue

drey: Ah, ok. Refactored. Thanks!

PR #549, opened 2026-01-27
feat: auto delete dependency detection

dmmordvi:
<div id='description'>
<h3>Summary by Bito</h3>

<p>This PR introduces automatic dependency detection for resource deletion in the nelm deployment system, enabling the system to automatically identify and manage internal dependencies for various Kubernetes resource types during deletion operations. The changes refactor the core dependency logic by removing hardcoded conditional parsing in service account dependency detection, improving modularity and allowing the system to handle dependencies more effectively without manual assignments. This enhances the deployment tool's reliability in managing complex Kubernetes resource lifecycles by automating what was previously explicit and simplifying plan building operations.</p>

<details>
<ul>

<li>Refactors parseServiceAccountDependencies in dependency.go to use a switch statement on GroupVersionKind().GroupKind() for determining nested paths to service account names, replacing hardcoded if-else conditions for different resource types like Pod, Deployment, DaemonSet, StatefulSet, ReplicaSet, ReplicationController, Job, and CronJob.</li>

<li>Modifies connectInternalDeployDependencies in plan_build.go to directly assign dependUponOps from findTrackAbsenceOpInStage and check length instead of using a found flag, adding a TODO comment for future multi-operation dependency handling in v2.</li>

<li>Simplifies connectInternalDeleteDependencies in plan_build.go by removing boolean variables and directly iterating over dependUponOps returned from findTrackAbsenceOpInStage to connect dependencies.</li>

<li>Changes findTrackAbsenceOpInStage in plan_build.go return type from ([]*Operation, bool) to []*Operation, updating logic to return nil on no matches or a single-operation slice, and removing redundant found checks.</li>

</ul>
</details>

</div>

### DIFF
--- a/internal/plan/plan_build.go
+++ b/internal/plan/plan_build.go
@@ -629,9 +629,11 @@ func connectInternalDeployDependencies(plan *Plan, instInfos []*InstallableResou
 			case common.ResourceStateReady:
 				dependUponOp, dependUponOpFound = findTrackReadinessOpInStage(plan, instInfos, dep, info.Stage)
 			case common.ResourceStateAbsent:
-				dependUponOps, dependUponOpFound := findTrackAbsenceOpInStage(plan, delInfos, instInfos, dep, info.Stage)
-				if dependUponOpFound && len(dependUponOps) > 0 {
+				// TODO(v2): all deploy/delete dependencies must depend upon all matched operations, not a single one
+				dependUponOps := findTrackAbsenceOpInStage(plan, delInfos, instInfos, dep, info.Stage)
+				if len(dependUponOps) > 0 {
 					dependUponOp = dependUponOps[0]
+					dependUponOpFound = true
 				}
 			default:
 				panic("unexpected internal dependency resource state")
@@ -683,20 +685,13 @@ func connectInternalDeleteDependencies(plan *Plan, delInfos []*DeletableResource
 		}
 
 		for _, dep := range internalDeps {
-			var (
-				dependUponOps      []*Operation
-				dependUponOpsFound bool
-			)
+			var dependUponOps []*Operation
 
 			switch dep.ResourceState {
 			case common.ResourceStateAbsent:
-				dependUponOps, dependUponOpsFound = findTrackAbsenceOpInStage(plan, delInfos, instInfos, dep, info.Stage)
+				dependUponOps = findTrackAbsenceOpInStage(plan, delInfos, instInfos, dep, info.Stage)
 			default:
 				panic("unexpected internal dependency resource state")
-			}
-
-			if !dependUponOpsFound {
-				continue
 			}
 
 			for _, dependUponOp := range dependUponOps {
@@ -762,7 +757,7 @@ func findTrackReadinessOpInStage(plan *Plan, instInfos []*InstallableResourceInf
 	return plan.Operation(opID)
 }
 
-func findTrackAbsenceOpInStage(plan *Plan, delInfos []*DeletableResourceInfo, instInfos []*InstallableResourceInfo, dep *resource.InternalDependency, sourceStage common.Stage) ([]*Operation, bool) {
+func findTrackAbsenceOpInStage(plan *Plan, delInfos []*DeletableResourceInfo, instInfos []*InstallableResourceInfo, dep *resource.InternalDependency, sourceStage common.Stage) []*Operation {
 	var foundOps []*Operation
 	for _, candidate := range delInfos {
 		if !candidate.MustTrackAbsence ||
@@ -779,7 +774,7 @@ func findTrackAbsenceOpInStage(plan *Plan, delInfos []*DeletableResourceInfo, in
 	}
 
 	if len(foundOps) > 0 {
-		return foundOps, len(foundOps) > 0
+		return foundOps
 	}
 
 	var match *InstallableResourceInfo
@@ -795,14 +790,15 @@ func findTrackAbsenceOpInStage(plan *Plan, delInfos []*DeletableResourceInfo, in
 	}
 
 	if match == nil {
-		return foundOps, false
+		return nil
 	}
 
 	opID := OperationID(OperationTypeTrackAbsence, OperationVersionTrackAbsence, OperationIteration(match.Iteration), match.ID())
 
-	if op, found := plan.Operation(opID); found {
-		foundOps = append(foundOps, op)
+	op, found := plan.Operation(opID)
+	if !found {
+		return nil
 	}
 
-	return foundOps, len(foundOps) > 0
+	return []*Operation{op}
 }
--- a/internal/resource/dependency.go
+++ b/internal/resource/dependency.go
@@ -689,19 +689,25 @@ func parseRoleDependencies(unstruct *unstructured.Unstructured, otherUnstructs [
 
 func parseServiceAccountDependencies(unstruct *unstructured.Unstructured, otherUnstructs []*unstructured.Unstructured) (dependencies []*InternalDependency, found bool) {
 	for _, otherUnstruct := range otherUnstructs {
-		var serviceAccountName string
-		// Get service account name for different resource kinds
-		if name, found := nestedString(otherUnstruct.Object, "spec", "serviceAccountName"); found { // Pod
-			serviceAccountName = name
-		} else if name, found = nestedString(otherUnstruct.Object, "spec", "template", "spec", "serviceAccountName"); found { // Controllers
-			serviceAccountName = name
-		} else if name, found = nestedString(otherUnstruct.Object, "spec", "jobTemplate", "spec", "template", "spec", "serviceAccountName"); found { // CronJob
-			serviceAccountName = name
-		} else {
+		var nestedPath []string
+		switch otherUnstruct.GroupVersionKind().GroupKind() {
+		case
+			schema.GroupKind{Kind: "Pod", Group: ""}:
+			nestedPath = []string{"spec", "serviceAccountName"}
+		case schema.GroupKind{Kind: "Deployment", Group: "apps"},
+			schema.GroupKind{Kind: "DaemonSet", Group: "apps"},
+			schema.GroupKind{Kind: "StatefulSet", Group: "apps"},
+			schema.GroupKind{Kind: "ReplicaSet", Group: "apps"},
+			schema.GroupKind{Kind: "ReplicationController", Group: ""},
+			schema.GroupKind{Kind: "Job", Group: "batch"}:
+			nestedPath = []string{"spec", "template", "spec", "serviceAccountName"}
+		case schema.GroupKind{Kind: "CronJob", Group: "batch"}:
+			nestedPath = []string{"spec", "jobTemplate", "spec", "template", "spec", "serviceAccountName"}
+		default:
 			continue
 		}
 
-		if serviceAccountName != unstruct.GetName() {
+		if name, ok := nestedString(otherUnstruct.Object, nestedPath...); !ok || name != unstruct.GetName() {
 			continue
 		}
 


internal/plan/plan_build.go
```diff
@@ -684,28 +684,27 @@ func connectInternalDeleteDependencies(plan *Plan, infos []*DeletableResourceInf
...
-				dependUponOp, dependUponOpFound = lo.Find(trackOps, func(op *Operation) bool {
+				dependUponOps = lo.Filter(trackOps, func(op *Operation, _ int) bool {
 					return dep.Match(getOpMeta(op))
 				})
 			default:
 				panic("unexpected internal dependency resource state")
 			}
 
-			if !dependUponOpFound {
+			if len(dependUponOps) == 0 {
```
ilya-lesikov: I guess this check is redundant

internal/resource/resource.go
```diff
@@ -170,40 +170,45 @@ type DeletableResourceOptions struct {
...
-	if err := validateDeleteDependencies(spec.ResourceMeta); err == nil {
-		manIntDeps = manualInternalDeleteDependencies(spec.ResourceMeta)
+	if err := validateDeleteDependencies(resourceSpec.ResourceMeta); err == nil {
+		manIntDeps = manualInternalDeleteDependencies(resourceSpec.ResourceMeta)
 	}
 
+	unstructList := lo.Map(otherResourceSpecs, func(itm *spec.ResourceSpec, _ int) *unstructured.Unstructured {
+		return itm.Unstruct
+	})
+	autoIntDeps := internalDeleteDependencies(resourceSpec.Unstruct, unstructList)
```
ilya-lesikov: I guess just inline it back for simplicity

internal/resource/resource.go
```diff
@@ -170,40 +170,45 @@ type DeletableResourceOptions struct {
...
 	}
 
 	var manIntDeps []*InternalDependency
-	if err := validateDeleteDependencies(spec.ResourceMeta); err == nil {
-		manIntDeps = manualInternalDeleteDependencies(spec.ResourceMeta)
+	if err := validateDeleteDependencies(resourceSpec.ResourceMeta); err == nil {
+		manIntDeps = manualInternalDeleteDependencies(resourceSpec.ResourceMeta)
 	}
 
+	unstructList := lo.Map(otherResourceSpecs, func(itm *spec.ResourceSpec, _ int) *unstructured.Unstructured {
```
ilya-lesikov: I'd say let's have a more specific naming: resSpec/rs, instead of itm

internal/resource/resource.go
```diff
@@ -218,7 +223,9 @@ type BuildResourcesOptions struct {
 func BuildResources(ctx context.Context, deployType common.DeployType, releaseNamespace string, prevRelResSpecs, newRelResSpecs []*spec.ResourceSpec, patchers []spec.ResourcePatcher, clientFactory kube.ClientFactorier, opts BuildResourcesOptions) ([]*InstallableResource, []*DeletableResource, error) {
 	var prevRelDelResources []*DeletableResource
 	for _, resSpec := range prevRelResSpecs {
-		deletableRes := NewDeletableResource(resSpec, releaseNamespace, DeletableResourceOptions{
+		otherResSpecs := lo.Without(prevRelResSpecs, resSpec)
```
ilya-lesikov: its probably will be more clear AND shorted if we inline it

internal/resource/resource_test.go
```diff
@@ -1287,7 +1295,9 @@ func runDeletableResourceTest(tc deletableResourceTestCase, s *DeletableResource
 
 		resSpec := tc.inputFunc()
 
-		res := resource.NewDeletableResource(resSpec, s.releaseNamespace, resource.DeletableResourceOptions{})
+		var otherResSpecs []*spec.ResourceSpec
+
+		res := resource.NewDeletableResource(resSpec, otherResSpecs, s.releaseNamespace, resource.DeletableResourceOptions{})
```
ilya-lesikov: a bit shorter
```suggestion
		res := resource.NewDeletableResource(resSpec, []*spec.ResourceSpec{}, s.releaseNamespace, resource.DeletableResourceOptions{})
```

internal/resource/dependency.go
```diff
@@ -102,8 +103,107 @@ func internalDeployDependencies(unstruct *unstructured.Unstructured) []*Internal
 	return dependencies
 }
 
-func internalDeleteDependencies(_ *unstructured.Unstructured) []*InternalDependency {
+func internalDeleteDependencies(unstruct *unstructured.Unstructured, otherUnstructs []*unstructured.Unstructured) []*InternalDependency {
+	gvk := unstruct.GroupVersionKind()
+	gk := gvk.GroupKind()
```
ilya-lesikov: inline both, no point
```suggestion
```

internal/resource/resource_test.go
ilya-lesikov: a bit shorter
```suggestion
		res := resource.NewDeletableResource(resSpec, []*spec.ResourceSpec{}, s.releaseNamespace, resource.DeletableResourceOptions{})
```

internal/resource/dependency.go
```diff
@@ -102,8 +103,107 @@ func internalDeployDependencies(unstruct *unstructured.Unstructured) []*Internal
...
+				Namespaces: []string{unstruct.GetName()},
+			},
+			ResourceState: common.ResourceStateAbsent,
+		}
+		dependencies = append(dependencies, dep)
+	case schema.GroupKind{Kind: "ClusterRole", Group: "rbac.authorization.k8s.io"}:
+		filtered := filterByGK(otherUnstructs,
+			schema.GroupKind{Kind: "RoleBinding", Group: "rbac.authorization.k8s.io"},
+			schema.GroupKind{Kind: "ClusterRoleBinding", Group: "rbac.authorization.k8s.io"},
+		)
```
ilya-lesikov: Let's stupidify and get rid of filterByGK, the logic is pretty simple:

```suggestion
filtered := return lo.Filter(otherUnstructs, func(unstruct *unstructured.Unstructured, _ int) bool {
	switch unstruct.GroupVersionKind().GroupKind():
	case schema.GroupKind{Kind: "RoleBinding", Group: "rbac.authorization.k8s.io"},
		schema.GroupKind{Kind: "ClusterRoleBinding", Group: "rbac.authorization.k8s.io"}:
		return true
	default:
		return false
	}
})
```

internal/resource/dependency.go
```diff
@@ -102,8 +103,107 @@ func internalDeployDependencies(unstruct *unstructured.Unstructured) []*Internal
...
+
+		for _, item := range filtered {
+			kind, found := nestedString(item.Object, "roleRef", "kind")
+
+			if !found || kind != "ClusterRole" {
```
ilya-lesikov: no space above

internal/resource/dependency.go
```diff
@@ -102,8 +103,107 @@ func internalDeployDependencies(unstruct *unstructured.Unstructured) []*Internal
...
+				continue
+			}
+
+			name, found := nestedString(item.Object, "roleRef", "name")
+
+			if !found || name != unstruct.GetName() {
```
ilya-lesikov: no space above

internal/resource/dependency.go
```diff
@@ -102,8 +103,107 @@ func internalDeployDependencies(unstruct *unstructured.Unstructured) []*Internal
...
+			ResourceState: common.ResourceStateAbsent,
+		}
+		dependencies = append(dependencies, dep)
+	case schema.GroupKind{Kind: "ClusterRole", Group: "rbac.authorization.k8s.io"}:
+		filtered := filterByGK(otherUnstructs,
+			schema.GroupKind{Kind: "RoleBinding", Group: "rbac.authorization.k8s.io"},
+			schema.GroupKind{Kind: "ClusterRoleBinding", Group: "rbac.authorization.k8s.io"},
+		)
+
+		for _, item := range filtered {
```
ilya-lesikov: improve naming: bindingUnstruct/bindingUnstructs instead of item/filtered

internal/resource/dependency.go
```diff
@@ -102,8 +103,107 @@ func internalDeployDependencies(unstruct *unstructured.Unstructured) []*Internal
...
+			dep := &InternalDependency{
+				ResourceMatcher: newExactResourceMatcher(item),
+				ResourceState:   common.ResourceStateAbsent,
+			}
+			dependencies = append(dependencies, dep)
+
+			if item.GetKind() == "ClusterRoleBinding" {
+				dep.ResourceMatcher.Namespaces = nil
+			}
+		}
```
ilya-lesikov: maybe more straightforward?

We can basically do this:
```suggestion
			var namespace string
			if item.GetKind() == "RoleBinding" {
				namespace = item.GetNamespace()
			}
			
			dep := &InternalDependency{
				ResourceMatcher: &spec.ResourceMatcher{
					Names:      []string{item.GetName()},
					Namespaces: namespace,
					Groups:     []string{item.GroupVersionKind().Group},
					Kinds:      []string{item.GetKind()},
				}
				ResourceState:   common.ResourceStateAbsent,
			}

			dependencies = append(dependencies, dep)
```

internal/resource/dependency.go
```diff
@@ -102,8 +103,107 @@ func internalDeployDependencies(unstruct *unstructured.Unstructured) []*Internal
...
+	case schema.GroupKind{Kind: "Role", Group: "rbac.authorization.k8s.io"}:
```
ilya-lesikov, line 117: same recommendations below

internal/resource/dependency.go
```diff
@@ -574,6 +674,57 @@ func parseRoleRef(unstruct unstructured.Unstructured) (dep *InternalDependency,
...
+	if !found {
+		return nil, false
+	}
+
+	// TODO(maybe): consider versions
+
+	dep = &InternalDependency{
+		ResourceMatcher: &spec.ResourceMatcher{
+			Names:      []string{},
+			Namespaces: []string{unstruct.GetNamespace()},
```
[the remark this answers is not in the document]
ilya-lesikov: Yes, we should never match for Namespace here. CRDs are clustered (no namespace). CRs can be clustered OR namespaced, and if namespaced they can have different namespaces too.

So the best and easisest thing is to not touch Namespace matcher at all — removing CRDs anyways affects all CRs, whatever the namespace is.

internal/resource/dependency.go
```diff
@@ -102,8 +103,107 @@ func internalDeployDependencies(unstruct *unstructured.Unstructured) []*Internal
...
+	gvk := unstruct.GroupVersionKind()
+	gk := gvk.GroupKind()
+
 	var dependencies []*InternalDependency
+	switch gk {
+	case schema.GroupKind{Kind: "CustomResourceDefinition", Group: "apiextensions.k8s.io"}:
+		if dep, found := parseCRD(unstruct); found {
+			dependencies = append(dependencies, dep)
+		}
+	case schema.GroupKind{Kind: "Namespace", Group: ""}:
```
ilya-lesikov: move code from each of these cases to the appropriate parseNamespace/... functions, like you did with parseCRD. For consistency and easier navigation in the switch

ilya-lesikov (changes requested):

internal/plan/plan_build.go
```diff
@@ -757,7 +762,8 @@ func findTrackReadinessOpInStage(plan *Plan, instInfos []*InstallableResourceInf
 	return plan.Operation(opID)
 }
 
-func findTrackAbsenceOpInStage(plan *Plan, delInfos []*DeletableResourceInfo, instInfos []*InstallableResourceInfo, dep *resource.InternalDependency, sourceStage common.Stage) (*Operation, bool) {
+func findTrackAbsenceOpInStage(plan *Plan, delInfos []*DeletableResourceInfo, instInfos []*InstallableResourceInfo, dep *resource.InternalDependency, sourceStage common.Stage) ([]*Operation, bool) {
```
ilya-lesikov: small codestyle moment: I do prefer to return found bool, if thing can be not found (like here), to make API clearer. Because usually you must return something, unless error, that's what users expect.

But that's for singular things, for plural things like slices its not needed, because most of the time user kinda expects that there might be 0 elem slice, and often it loops over it anyway, so it won't break even if empty. 

And yea, even the code existing before this PR in dependency.go could probably be simplified in the same way, but whatever. If `l := append(l, nil...)` can work.

internal/plan/plan_build.go
```diff
@@ -767,7 +773,13 @@ func findTrackAbsenceOpInStage(plan *Plan, delInfos []*DeletableResourceInfo, in
...
 		opID := OperationID(OperationTypeTrackAbsence, OperationVersionTrackAbsence, 0, candidate.ID())
 
-		return plan.Operation(opID)
+		if op, found := plan.Operation(opID); found {
+			foundOps = append(foundOps, op)
+		}
+	}
+
+	if len(foundOps) > 0 {
+		return foundOps, len(foundOps) > 0
```
ilya-lesikov: could be more explicit: `return foundOps, true`

internal/plan/plan_build.go
```diff
@@ -783,10 +795,14 @@ func findTrackAbsenceOpInStage(plan *Plan, delInfos []*DeletableResourceInfo, in
 	}
 
 	if match == nil {
-		return nil, false
+		return foundOps, false
```
ilya-lesikov: revert to keep it explicit

internal/plan/plan_build.go
```diff
@@ -783,10 +795,14 @@ func findTrackAbsenceOpInStage(plan *Plan, delInfos []*DeletableResourceInfo, in
...
 	}
 
 	opID := OperationID(OperationTypeTrackAbsence, OperationVersionTrackAbsence, OperationIteration(match.Iteration), match.ID())
 
-	return plan.Operation(opID)
+	if op, found := plan.Operation(opID); found {
+		foundOps = append(foundOps, op)
+	}
+
+	return foundOps, len(foundOps) > 0
```
ilya-lesikov: Could even do early return, its fine though
```suggestion
	op, found := plan.Operation(opID)
	if !found {
		return nil, false 
	}

	return []*Operation{op}, true

```

internal/plan/plan_build.go
```diff
@@ -629,7 +629,10 @@ func connectInternalDeployDependencies(plan *Plan, instInfos []*InstallableResou
 			case common.ResourceStateReady:
 				dependUponOp, dependUponOpFound = findTrackReadinessOpInStage(plan, instInfos, dep, info.Stage)
 			case common.ResourceStateAbsent:
-				dependUponOp, dependUponOpFound = findTrackAbsenceOpInStage(plan, delInfos, instInfos, dep, info.Stage)
+				dependUponOps, dependUponOpFound := findTrackAbsenceOpInStage(plan, delInfos, instInfos, dep, info.Stage)
```
ilya-lesikov: I think this should stay like that until v2. Let's add a comment above this line:
```
// TODO(v2): all deploy/delete dependencies must depend upon all matched operations, not a single one
```

internal/resource/dependency.go
```diff
@@ -574,6 +597,159 @@ func parseRoleRef(unstruct unstructured.Unstructured) (dep *InternalDependency,
...
+		// Get service account name for different resource kinds
+		if name, found := nestedString(otherUnstruct.Object, "spec", "serviceAccountName"); found { // Pod
+			serviceAccountName = name
+		} else if name, found = nestedString(otherUnstruct.Object, "spec", "template", "spec", "serviceAccountName"); found { // Controllers
+			serviceAccountName = name
+		} else if name, found = nestedString(otherUnstruct.Object, "spec", "jobTemplate", "spec", "template", "spec", "serviceAccountName"); found { // CronJob
+			serviceAccountName = name
+		} else {
+			continue
+		}
```
ilya-lesikov: You know, that's a bit too nuclear. Especially spec.serviceAccountName. This might match CRDs we know nothing about.

Let's keep it safe and match exact Pod, CronJob, Job, Deployment, StatefulSet, DaemonSet, ReplicaSet, ReplicationController controllers (did I miss anything?). You can grep this file, they are all should be there somewhere, to figure out how to match them.

And then for each of them pick the field they actually use, without these heuristics.

ilya-lesikov (changes requested):

PR #564, opened 2026-02-12
feat: add plan freezing support

drey:
<div id='description'>
<h3>Summary by Bito</h3>

<p>This PR introduces plan freezing support, enabling users to save release install plans as encrypted artifacts and execute them later for offline deployments or approval workflows. The implementation removes the feature gate to enable the functionality by default, updates CLI commands with new flags for plan artifact operations, adds JSON serialization capabilities to core plan structures, and modifies action handlers to support artifact-based plan execution and display. Additionally, it refactors OperationConfig structs to use JSON struct tags for serialization instead of custom MarshalJSON methods, enabling plan freezing by allowing serialization of operation configurations.</p>

<details>
<ul>

<li>Introduces plan freezing functionality by removing the FeatGatePlanFreezing feature gate in feat.go and implementing PlanArtifact structure with encryption and validation in internal/plan/plan_artifact.go.</li>

<li>Refactors OperationConfig structs to use JSON struct tags instead of custom MarshalJSON methods in internal/plan/operation_config.go, enabling serialization of operation configurations.</li>

<li>Adds new CLI commands 'release plan install' and 'release plan show' in cmd/nelm/ and modifies the 'release install' command with --use-plan and --plan-lifetime flags for plan artifact operations.</li>

<li>Updates action handlers in pkg/action/ to integrate plan artifact reading, validation, and execution, while renaming structs like ResourceChangeUDiffOptions to ResourceDiffOptions in pkg/common/options.go.</li>

</ul>
</details>

</div>

### DIFF
--- a/cmd/nelm/release_install.go
+++ b/cmd/nelm/release_install.go
@@ -249,7 +249,7 @@ func newReleaseInstallCommand(ctx context.Context, afterAllCommandsBuiltFuncs ma
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(ilya-lesikov): restrict allowed values
+		// TODO: restrict allowed values
 		if err := cli.AddFlag(cmd, &cfg.ReleaseStorageDriver, "release-storage", "", "How releases should be stored", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalEnvVarRegexes,
 			Group:                miscFlagGroup,
--- a/internal/plan/resource_info.go
+++ b/internal/plan/resource_info.go
@@ -154,7 +154,7 @@ func BuildResourceInfos(ctx context.Context, deployType common.DeployType, relea
 	return instResourceInfos, delResourceInfos, nil
 }
 
-// TODO(v2): keep annotation should probably forbid resource recreations
+// TODO(major): keep annotation should probably forbid resource recreations
 func buildInstallableResourceInfo(ctx context.Context, localRes *resource.InstallableResource, deployType common.DeployType, releaseNamespace string, prevRelFailed, noRemoveManualChanges bool, clientFactory kube.ClientFactorier) ([]*InstallableResourceInfo, error) {
 	var stages []common.Stage
 	switch deployType {
--- a/internal/resource/spec/resource_spec.go
+++ b/internal/resource/spec/resource_spec.go
@@ -25,7 +25,7 @@ type ResourceSpec struct {
 
 type ResourceSpecOptions struct {
 	FilePath                string
-	LegacyNoCleanNullFields bool // TODO(v2): always clean
+	LegacyNoCleanNullFields bool // TODO(major): always clean
 	StoreAs                 common.StoreAs
 }
 
@@ -129,7 +129,7 @@ func BuildReleasableResourceSpecs(ctx context.Context, releaseNamespace string, 
 		for _, resPatcher := range patchers {
 			if matched, err := resPatcher.Match(ctx, &ResourcePatcherResourceInfo{
 				Obj: releasableRes.Unstruct,
-				// TODO(ilya-lesikov): get rid of ownership for releasable resources
+				// TODO: get rid of ownership for releasable resources
 				Ownership: "",
 			}); err != nil {
 				return nil, fmt.Errorf("match resource for patching by %q: %w", resPatcher.Type(), err)
@@ -147,7 +147,7 @@ func BuildReleasableResourceSpecs(ctx context.Context, releaseNamespace string, 
 
 			patchedObj, err := resPatcher.Patch(ctx, &ResourcePatcherResourceInfo{
 				Obj: unstruct,
-				// TODO(ilya-lesikov): get rid of ownership for releasable resources
+				// TODO: get rid of ownership for releasable resources
 				Ownership: "",
 			})
 			if err != nil {
--- a/pkg/action/release_install.go
+++ b/pkg/action/release_install.go
@@ -59,7 +59,7 @@ type ReleaseInstallOptions struct {
 	// Used to set application version metadata without modifying the chart file.
 	ChartAppVersion string
 	// ChartDirPath is deprecated
-	ChartDirPath string // TODO(v2): get rid
+	ChartDirPath string // TODO(major): get rid
 	// ChartProvenanceKeyring is the path to a keyring file containing public keys
 	// used to verify chart provenance signatures. Used with signed charts for security.
 	ChartProvenanceKeyring string
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -39,15 +39,15 @@ var ChartTSEntryPoints = [...]string{ChartTSEntryPointTS, ChartTSEntryPointJS}
 
 const (
 	DefaultBurstLimit = 100
-	// TODO(v2): switch to if-possible
+	// TODO(major): switch to if-possible
 	DefaultChartProvenanceStrategy = "never"
-	// TODO(v2): reconsider?
+	// TODO(major): reconsider?
 	DefaultDeletePropagation = metav1.DeletePropagationForeground
 	DefaultDiffContextLines  = 3
 	DefaultFieldManager      = "helm"
 	// DefaultResourceValidationKubeVersion Kubernetes version to use during resource validation by kubeconform
 	DefaultResourceValidationKubeVersion = "1.35.0"
-	// TODO(v2): update to a more recent version? Not sure about backwards compatibility.
+	// TODO(major): update to a more recent version? Not sure about backwards compatibility.
 	DefaultLocalKubeVersion      = "1.20.0"
 	DefaultLogColorMode          = log.LogColorModeAuto
 	DefaultNetworkParallelism    = 30
@@ -283,7 +283,7 @@ var (
 	AnnotationKeyHumanDeleteDependency   = "werf.io/delete-dependency-<name>"
 	AnnotationKeyPatternDeleteDependency = regexp.MustCompile(`^werf.io/delete-dependency-(?P<id>.+)$`)
 
-	// TODO(v2): get rid
+	// TODO(major): get rid
 	AnnotationKeyHumanDependency   = "<name>.dependency.werf.io"
 	AnnotationKeyPatternDependency = regexp.MustCompile(`^(?P<id>.+).dependency.werf.io$`)
 
--- a/pkg/common/options.go
+++ b/pkg/common/options.go
@@ -127,7 +127,7 @@ type ValuesOptions struct {
 	// RuntimeSetJSON is a list of key-value pairs in "key=json" format to set in $.Runtime.
 	// This is meant to be generated programmatically. Users should prefer ValuesSetJSON.
 	// Example: ["runtime.env=dev", "runtime.timestamp=1234567890"]
-	// TODO(v2): get rid of it
+	// TODO(major): get rid of it
 	RuntimeSetJSON []string
 	// ValuesFiles is a list of paths to additional values files to merge with chart values.
 	// Files are merged in order, with later files overriding earlier ones.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,7 +17,7 @@ These are NOT suggestions — they are requirements. NEVER skip them in favor of
 
 These are NOT suggestions — they are requirements. NEVER skip them in favor of Grep.
 
-- ALWAYS use **LSP** `hover` to look up documentation, type signatures, and godoc for any symbol in the codebase (including symbols from external dependencies).
+- ALWAYS use **LSP** `hover` to look up documentation, type signatures, and godoc for any symbol in the codebase (including symbols from external dependencies)
 - ALWAYS use **Context7 MCP** (`resolve-library-id` + `query-docs`) to look up broader documentation, guides, or examples for any technology, library, or tool. NEVER guess at APIs — look them up.
 
 ## Verifying changes (MANDATORY)
--- a/cmd/nelm/chart_lint.go
+++ b/cmd/nelm/chart_lint.go
@@ -225,7 +225,7 @@ func newChartLintCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*co
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(ilya-lesikov): restrict allowed values
+		// TODO: restrict allowed values
 		if err := cli.AddFlag(cmd, &cfg.ReleaseStorageDriver, "release-storage", "", "How releases should be stored", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalEnvVarRegexes,
 			Group:                miscFlagGroup,
--- a/cmd/nelm/chart_render.go
+++ b/cmd/nelm/chart_render.go
@@ -200,7 +200,7 @@ func newChartRenderCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(ilya-lesikov): restrict allowed values
+		// TODO: restrict allowed values
 		if err := cli.AddFlag(cmd, &cfg.ReleaseStorageDriver, "release-storage", "", "How releases should be stored", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalEnvVarRegexes,
 			Group:                miscFlagGroup,
--- a/cmd/nelm/common_flags.go
+++ b/cmd/nelm/common_flags.go
@@ -388,7 +388,7 @@ func AddValuesFlags(cmd *cobra.Command, cfg *common.ValuesOptions) error {
 		return fmt.Errorf("add flag: %w", err)
 	}
 
-	// TODO(v2): revise all flags in nelm/werf to make sure they are all parsed as it happens in
+	// TODO(major): revise all flags in nelm/werf to make sure they are all parsed as it happens in
 	// Helm (see https://github.com/werf/nelm/issues/337)
 	if err := cli.AddFlag(cmd, &cfg.ValuesSet, "set", []string{}, "Set new values, where the key is the value path and the value is the value", cli.AddFlagOptions{
 		GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
--- a/cmd/nelm/release_get.go
+++ b/cmd/nelm/release_get.go
@@ -71,7 +71,7 @@ func newReleaseGetCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*c
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(ilya-lesikov): restrict values
+		// TODO: restrict values
 		if err := cli.AddFlag(cmd, &cfg.OutputFormat, "output-format", action.DefaultReleaseGetOutputFormat, "Result output format", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
 			Group:                miscFlagGroup,
@@ -86,7 +86,7 @@ func newReleaseGetCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*c
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(ilya-lesikov): restrict allowed values
+		// TODO: restrict allowed values
 		if err := cli.AddFlag(cmd, &cfg.ReleaseStorageDriver, "release-storage", "", "How releases should be stored", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalEnvVarRegexes,
 			Group:                miscFlagGroup,
--- a/cmd/nelm/release_list.go
+++ b/cmd/nelm/release_list.go
@@ -57,7 +57,7 @@ func newReleaseListCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(ilya-lesikov): restrict values
+		// TODO: restrict values
 		if err := cli.AddFlag(cmd, &cfg.OutputFormat, "output-format", action.DefaultReleaseListOutputFormat, "Result output format", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
 			Group:                miscFlagGroup,
@@ -73,7 +73,7 @@ func newReleaseListCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(ilya-lesikov): restrict allowed values
+		// TODO: restrict allowed values
 		if err := cli.AddFlag(cmd, &cfg.ReleaseStorageDriver, "release-storage", "", "How releases should be stored", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalEnvVarRegexes,
 			Group:                miscFlagGroup,
--- a/cmd/nelm/release_plan_install.go
+++ b/cmd/nelm/release_plan_install.go
@@ -246,7 +246,7 @@ func newReleasePlanInstallCommand(ctx context.Context, afterAllCommandsBuiltFunc
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(ilya-lesikov): restrict allowed values
+		// TODO: restrict allowed values
 		if err := cli.AddFlag(cmd, &cfg.ReleaseStorageDriver, "release-storage", "", "How releases should be stored", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalEnvVarRegexes,
 			Group:                miscFlagGroup,
@@ -282,7 +282,7 @@ func newReleasePlanInstallCommand(ctx context.Context, afterAllCommandsBuiltFunc
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(v2): get rid?
+		// TODO(major): get rid?
 		if err := cli.AddFlag(cmd, &cfg.ShowVerboseDiffs, "show-verbose-diffs", true, "Show verbose diff lines", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagLocalEnvVarRegexes,
 			Group:                mainFlagGroup,
--- a/cmd/nelm/release_rollback.go
+++ b/cmd/nelm/release_rollback.go
@@ -141,7 +141,7 @@ func newReleaseRollbackCommand(ctx context.Context, afterAllCommandsBuiltFuncs m
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(ilya-lesikov): restrict allowed values
+		// TODO: restrict allowed values
 		if err := cli.AddFlag(cmd, &cfg.ReleaseStorageDriver, "release-storage", "", "How releases should be stored", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalEnvVarRegexes,
 			Group:                miscFlagGroup,
@@ -156,7 +156,7 @@ func newReleaseRollbackCommand(ctx context.Context, afterAllCommandsBuiltFuncs m
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(v2): remove this duplicated flag
+		// TODO(major): remove this duplicated flag
 		if err := cli.AddFlag(cmd, &cfg.RollbackGraphPath, "save-rollback-graph-to", "", "Save the Graphviz rollback graph to a file", cli.AddFlagOptions{
 			Group: mainFlagGroup,
 			Type:  cli.FlagTypeFile,
--- a/cmd/nelm/release_uninstall.go
+++ b/cmd/nelm/release_uninstall.go
@@ -90,7 +90,7 @@ func newReleaseUninstallCommand(ctx context.Context, afterAllCommandsBuiltFuncs 
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(ilya-lesikov): restrict allowed values
+		// TODO: restrict allowed values
 		if err := cli.AddFlag(cmd, &cfg.ReleaseStorageDriver, "release-storage", "", "How releases should be stored", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalEnvVarRegexes,
 			Group:                miscFlagGroup,
--- a/cmd/nelm/release_uninstall_legacy.go
+++ b/cmd/nelm/release_uninstall_legacy.go
@@ -113,7 +113,7 @@ func newLegacyReleaseUninstallCommand(ctx context.Context, afterAllCommandsBuilt
 			return fmt.Errorf("add flag: %w", err)
 		}
 
-		// TODO(ilya-lesikov): restrict allowed values
+		// TODO: restrict allowed values
 		if err := cli.AddFlag(cmd, &cfg.ReleaseStorageDriver, "release-storage", "", "How releases should be stored", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalEnvVarRegexes,
 			Group:                miscFlagGroup,
--- a/internal/chart/chart_download.go
+++ b/internal/chart/chart_download.go
@@ -49,9 +49,9 @@ func newChartDownloader(ctx context.Context, chartRef string, registryClient *he
 			helmgetter.WithTimeout(opts.ChartRepoRequestTimeout),
 		},
 		RegistryClient: registryClient,
-		// TODO(v2): get rid of HELM_ env vars support
+		// TODO(major): get rid of HELM_ env vars support
 		RepositoryConfig: cli.EnvOr("HELM_REPOSITORY_CONFIG", helmpath.ConfigPath("repositories.yaml")),
-		// TODO(v2): get rid of HELM_ env vars support
+		// TODO(major): get rid of HELM_ env vars support
 		RepositoryCache: cli.EnvOr("HELM_REPOSITORY_CACHE", helmpath.CachePath("repository")),
 	}
 
@@ -97,12 +97,12 @@ func downloadChart(ctx context.Context, chartPath string, registryClient *helmre
 			return "", fmt.Errorf("construct chart downloader: %w", err)
 		}
 
-		// TODO(v2): get rid of HELM_ env vars support
+		// TODO(major): get rid of HELM_ env vars support
 		if err := os.MkdirAll(cli.EnvOr("HELM_REPOSITORY_CACHE", helmpath.CachePath("repository")), 0o755); err != nil {
 			return "", fmt.Errorf("create repository cache directory: %w", err)
 		}
 
-		// TODO(v2): get rid of HELM_ env vars support
+		// TODO(major): get rid of HELM_ env vars support
 		chartPath, _, err = chartDownloader.DownloadTo(chartRef, opts.ChartVersion, cli.EnvOr("HELM_REPOSITORY_CACHE", helmpath.CachePath("repository")))
 		if err != nil {
 			return "", fmt.Errorf("download chart %q: %w", chartRef, err)
--- a/internal/chart/chart_render.go
+++ b/internal/chart/chart_render.go
@@ -80,9 +80,9 @@ func RenderChart(ctx context.Context, chartPath, releaseName, releaseNamespace s
 		SkipUpdate:     opts.ChartRepoNoUpdate,
 		Getters:        getter.Providers{getter.HttpProvider, getter.OCIProvider},
 		RegistryClient: registryClient,
-		// TODO(v2): don't read HELM_REPOSITORY_CONFIG anymore
+		// TODO(major): don't read HELM_REPOSITORY_CONFIG anymore
 		RepositoryConfig: cli.EnvOr("HELM_REPOSITORY_CONFIG", helmpath.ConfigPath("repositories.yaml")),
-		// TODO(v2): don't read HELM_REPOSITORY_CACHE anymore
+		// TODO(major): don't read HELM_REPOSITORY_CACHE anymore
 		RepositoryCache:   cli.EnvOr("HELM_REPOSITORY_CACHE", helmpath.CachePath("repository")),
 		AllowMissingRepos: true,
 	}
--- a/internal/kube/legacy_client_getter.go
+++ b/internal/kube/legacy_client_getter.go
@@ -17,7 +17,7 @@ type LegacyClientGetter struct {
 	legacyClientConfig clientcmd.ClientConfig
 }
 
-// TODO(v2): get rid
+// TODO(major): get rid
 func NewLegacyClientGetter(discoveryClient discovery.CachedDiscoveryInterface, mapper meta.ResettableRESTMapper, restConfig *rest.Config, legacyClientConfig clientcmd.ClientConfig) *LegacyClientGetter {
 	return &LegacyClientGetter{
 		discoveryClient:    discoveryClient,
--- a/internal/legacy/deploy/resources_waiter.go
+++ b/internal/legacy/deploy/resources_waiter.go
@@ -66,7 +66,7 @@ func init() {
 	flaggerscheme.AddToScheme(scheme.Scheme)
 }
 
-// TODO(v2): get rid
+// TODO(major): get rid
 type ResourcesWaiter struct {
 	Client                    *helm_kube.Client
 	LogsFromTime              time.Time
--- a/internal/legacy/deploy/stages_splitter.go
+++ b/internal/legacy/deploy/stages_splitter.go
@@ -18,7 +18,7 @@ func NewStagesSplitter() *StagesSplitter {
 	return &StagesSplitter{}
 }
 
-// TODO(v2): get rid
+// TODO(major): get rid
 type StagesSplitter struct{}
 
 func (s *StagesSplitter) Split(resources kube.ResourceList) (stages.SortedStageList, error) {
--- a/internal/lock/lock_manager.go
+++ b/internal/lock/lock_manager.go
@@ -108,7 +108,6 @@ func (lockManager *LockManager) LockRelease(
 	ctx context.Context,
 	releaseName string,
 ) (lockgate.LockHandle, error) {
-	// TODO: add support of context into lockgate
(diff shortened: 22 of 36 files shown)


cmd/nelm/release_plan_show.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func newReleasePlanShowCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*cobra.Command]func(cmd *cobra.Command) error) *cobra.Command {
+	cfg := &releasePlanShowConfig{}
+
+	cmd := cli.NewSubCommand(
+		ctx,
+		"show [options...] plan.json",
+		"Show plan artifact planned changes.",
+		"Show plan artifact planned changes.",
```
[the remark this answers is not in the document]
drey: This is not an error. The same phrase for both short and long description.

pkg/common/common.go
```diff
@@ -322,4 +322,7 @@ var (
 	DefaultResourceValidationCacheLifetime = 48 * time.Hour
 
 	APIResourceValidationJSONSchemasCacheDir = helmpath.CachePath("nelm", "api-resource-json-schemas")
+
+	// TODO: need to add CLI argument to configure this param on plan execution
+	APIPlanArtifactLifetime = 2 * time.Hour
```
ilya-lesikov: just DefaultPlanArtifactLifetime?

cmd/nelm/release_plan.go
```diff
@@ -20,5 +21,9 @@ func newPlanCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*cobra.C
 
 	cmd.AddCommand(newReleasePlanInstallCommand(ctx, afterAllCommandsBuiltFuncs))
 
+	if featgate.FeatGatePlanFreezing.Enabled() {
```
ilya-lesikov: Update description above of "nelm plan" category to something please

internal/plan/operation.go
```diff
@@ -101,3 +102,113 @@ func OperationIDHuman(t OperationType, iteration OperationIteration, configIDHum
...
 }
+
+type operationJSON struct {
+	Type      OperationType          `json:"type"`
+	Version   OperationVersion       `json:"version"`
+	Category  OperationCategory      `json:"category"`
+	Iteration OperationIteration     `json:"iteration"`
+	Status    OperationStatus        `json:"status"`
+	Config    operationConfigPayload `json:"config"`
+}
```
ilya-lesikov: Can't we avoid these extra structs? I specifically did structs like Operation in a way that they can be marshalled. Unless there are some issues with it?

internal/plan/operation_config.go
```diff
@@ -46,9 +51,29 @@ func (c *OperationConfigNoop) IDHuman() string {
...
 }
 
+func (c *OperationConfigNoop) Kind() string {
+	return string(OperationTypeNoop)
+}
+
+func (c *OperationConfigNoop) MarshalJSON() ([]byte, error) {
+	type operationConfigNoopJSON OperationConfigNoop
+
+	return json.Marshal((*operationConfigNoopJSON)(c)) //nolint:wrapcheck
```
ilya-lesikov: let's just wrap an error

internal/plan/operation_config.go
```diff
@@ -110,22 +215,47 @@ func (c *OperationConfigDelete) IDHuman() string {
...
-	SaveLogsOnlyForContainers                []string
-	SaveLogsOnlyForNumberOfReplicas          int
+	ResourceMeta *spec.ResourceMeta `json:"resourceMeta"`
+
+	FailMode                                 multitrack.FailMode       `json:"failMode"`
+	FailuresAllowed                          int                       `json:"failuresAllowed"`
+	IgnoreLogs                               bool                      `json:"ignoreLogs"`
+	IgnoreLogsForContainers                  []string                  `json:"ignoreLogsForContainers,omitempty"`
+	IgnoreLogsByRegex                        *regexp.Regexp            `json:"-"`
+	IgnoreLogsByRegexForContainers           map[string]*regexp.Regexp `json:"-"`
```
ilya-lesikov: difficult to marshal? Kinda don't want to lose it

ilya-lesikov: oh, I see now. Yea, kinda ugly...

ilya-lesikov: Maybe let's make a wrapper around Regex with marshal methods? Place it somewhere in internal/util. Will eliminate all of this complexity

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+	return a.Data.InstallableResourceInfos
+}
+
+func (a *PlanArtifact) GetPlan() *Plan {
+	return a.Data.DAG
+}
+
+func (a *PlanArtifact) GetRelease() (*release.Release, error) {
+	for _, op := range a.Data.DAG.Operations() {
+		if c, ok := op.Config.(*OperationConfigCreateRelease); ok {
```
ilya-lesikov: Nah, nothing should be extracted from the Plan. The plan can only be marshalled/unmarshalled, nothing else. Pack the whole new release in PlanArtifact separately.

Getter will be useless after that, too.

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+	Changes                  []*ResourceChange          `json:"changes"`
+	DAG                      *Plan                      `json:"dag"`
+	InstallableResourceInfos []*InstallableResourceInfo `json:"installableResourceInfos"`
+	ReleaseInfos             []*ReleaseInfo             `json:"releaseInfos"`
+}
+
+type PlanArtifactRelease struct {
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+	Version   int    `json:"version"`
```
ilya-lesikov: Revision instead of Version

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+	if err := createPlanArtifactArchive(ctx, file, artifact, secretKey, secretWorkDir); err != nil {
+		return fmt.Errorf("marshal plan artifact: %w", err)
+	}
+
+	return nil
+}
+
+func BuildPlanArtifact(p *Plan, deployType common.DeployType, changes []*ResourceChange, release PlanArtifactRelease, instResInfos []*InstallableResourceInfo, relInfos []*ReleaseInfo, opts PlanArtifactOptions) *PlanArtifact {
+	sort.Slice(changes, func(i, j int) bool {
+		return changes[i].ResourceMeta.IDHuman() < changes[j].ResourceMeta.IDHuman()
```
ilya-lesikov: Better use ID() for such things. IDHuman solely for presentation

There is also ResourceMetaSortHandler and ResourceSpecSortHandler. Kinda not easy to find unless you know what you are looking for, but oh well

Just saying. You don't need this sorting anyways

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+	defer file.Close()
+
+	if err := createPlanArtifactArchive(ctx, file, artifact, secretKey, secretWorkDir); err != nil {
+		return fmt.Errorf("marshal plan artifact: %w", err)
+	}
+
+	return nil
+}
+
+func BuildPlanArtifact(p *Plan, deployType common.DeployType, changes []*ResourceChange, release PlanArtifactRelease, instResInfos []*InstallableResourceInfo, relInfos []*ReleaseInfo, opts PlanArtifactOptions) *PlanArtifact {
```
ilya-lesikov: How about taking all code from this function and just inline it in release_plan_install.go? Otherwise you need to pass this gigantic number of args through the whole stack without really doing anything with them, doesn't make it simplier. And I think whatever comes out of CalculatePlannedChanges is already sorted, so no need for that.

pkg/action/release_plan_execute.go
```diff
@@ -0,0 +1,242 @@
...
+	"github.com/werf/nelm/internal/release"
+	"github.com/werf/nelm/pkg/common"
+	"github.com/werf/nelm/pkg/log"
+)
+
+const (
+	DefaultReleasePlanExecuteLogLevel = log.InfoLevel
+)
+
+type ReleasePlanExecuteOptions struct {
```
ilya-lesikov: Can we DRY it out somewhat? I'm thinking about this:
1. Only release_install.go file is left
2. Rename ReleasePlanExecuteOptions to ReleaseInstallRuntimeOptions 
3. Embed ReleaseInstallRuntimeOptions into ReleaseInstallOptions and ReleasePlanInstallOptions.
4. Bundle ReleaseInstallRuntimeOptions into the plan artifact
5. When loading plan artifact on --use-plan, replace original RuntimeOptions with baked in RuntimeOptions. This way you can continue using ReleaseInstallOptions in the same manner, whether --use-plan specified or not.
6. Code in ReleaseInstall() before PlanExecute() sometimes is shared, sometimes is guarded by "if (not) --use-plan".
7. Code starting from PlanExecute() and below should stay the same, 1:1.

Yes, we kinda change the ReleaseInstall API a bit with this change, but it's fine. Currently its only used in werf/deckhouse/marketplace.

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+	return a.DeployType
+}
+
+func (a *PlanArtifact) GetInstallableResourceInfos() []*InstallableResourceInfo {
+	return a.Data.InstallableResourceInfos
+}
+
+func (a *PlanArtifact) GetPlan() *Plan {
+	return a.Data.DAG
+}
```
ilya-lesikov: These getters useless, let's go without

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+			return c.Release, nil
+		}
+	}
+
+	return nil, fmt.Errorf("no create release operation found in plan artifact")
+}
+
+func (a *PlanArtifact) GetReleaseInfos() []*ReleaseInfo {
+	return a.Data.ReleaseInfos
+}
```
ilya-lesikov, line 31: One more useless getter

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	artifact.Data = &data
+
+	return &artifact, nil
+}
+
+func ValidatePlanArtifact(artifact *PlanArtifact, expectedRevision int, lifetime time.Duration) error {
+	if artifact == nil {
+		return fmt.Errorf("%w: plan is empty", ErrPlanArtifactInvalid)
```
ilya-lesikov: Let's make it in this style:
```suggestion
		return fmt.Errorf("plan shouldn't be empty")
```
All errors below too. Reasoning:
1. ErrPlanArtifactInvalid doesn't provide any extra context — failure in ValidatePlanArtifact is a validation failure by definition
2. We don't check anywhere for this error class now, so currently it is useless

While it might be a good idea in general to make typed errors throughout the codebase, still — so far we only needed error types like 2 (?) times in the whole codebase. So far it wasn't really worth it, so let's create typed errors only if there is an immediate case for this.

I'm not satisfied with errors in go in general. The current approach works ok, but has its issues. At some point we need to brainstorm error handling, maybe look for some good libraries/approaches? Or maybe just wait until Go developers improve it? It lacks many things, e.g. a lot of boilerplate, unstructured errors, etc.

internal/plan/planned_changes.go
```diff
@@ -32,13 +32,63 @@ type CalculatePlannedChangesOptions struct {
...
-	Type         string
-	TypeStyle    color.Style
-	Udiff        string
+	Type         string                     `json:"type"`
+	TypeStyle    color.Style                `json:"typeStyle"`
+	Before       *unstructured.Unstructured `json:"before"`
+	After        *unstructured.Unstructured `json:"after"`
+}
+
+func (c *ResourceChange) GetUDiff(opts CalculatePlannedChangesOptions) (string, error) {
```
ilya-lesikov: Go convention is UDiff() for getters and SetUDiff() for setters, let's follow it where we can

ilya-lesikov: Let's rename CalculatePlannedChangesOptions to ResourceChangeUDiffOptions. And remove it from CalcualtePlannedChanges and all other functions, where it's not needed anymore.

internal/plan/resource_info.go
```diff
@@ -53,21 +53,21 @@ func ResourceInstallTypeSortHandler(type1, type2 ResourceInstallType) bool {
...
+	*spec.ResourceMeta `json:"resourceMeta"`
 
-	LocalResource  *resource.InstallableResource
-	GetResult      *unstructured.Unstructured
-	DryApplyResult *unstructured.Unstructured
-	DryApplyErr    error
+	LocalResource  *resource.InstallableResource `json:"localResource"`
+	GetResult      *unstructured.Unstructured    `json:"-"`
+	DryApplyResult *unstructured.Unstructured    `json:"-"`
+	DryApplyErr    error                         `json:"-"`
```
ilya-lesikov: Is there a reason to exclude these? Just dump it and don't bother whether it will be useful down the line or not. Easier to maintain, less things to think/worry about. When it's just "we cut the release install in half and fully serialized all things needed to continue" it's easy to remember/understand/explain. But if that's "we kinda cut the release install and then did 20 more things that you need to think about everytime you touch any of this", it's not so fun anymore. 

And yeah, size is not an issue with gzip, its probably effectively deduplicates all of these unstructured things, so don't worry about it.

internal/plan/release_info.go
```diff
@@ -29,10 +29,10 @@ const (
 // Data class, which stores all info to make a decision on what to do with the release revision
 // in the plan.
 type ReleaseInfo struct {
-	Release *helmrelease.Release
+	Release *helmrelease.Release `json:"release"`
```
ilya-lesikov, line 32: Note that Files from Release.Chart will be dropped on serialization. We drop them because they can be kinda stupid big if TS is used in the chart or if there is something big in the charts, like some binaries. Probably won't affect us?

Also Release.UnstoredManifests are dropped, along with Release.Chart.SecretsRuntimeData and Release.Chart.ExtraValues

internal/resource/spec/resource_spec.go
```diff
@@ -17,10 +17,10 @@ import (
 // Enough to create or update resources, as well as delete, get, etc. Use it when ResourceMeta is
 // not enough and you also need the actual resource spec.
 type ResourceSpec struct {
-	*ResourceMeta
+	*ResourceMeta `json:"meta"`
```
ilya-lesikov: since in other places its resourceMeta, let's make it here too

internal/resource/resource.go
```diff
@@ -21,35 +22,40 @@ import (
...
+	AutoInternalDependencies               []*InternalDependency           `json:"autoInternalDependencies,omitempty"`
+	ExternalDependencies                   []*ExternalDependency           `json:"externalDependencies,omitempty"`
+	DeployConditions                       map[common.On][]common.Stage    `json:"deployConditions"`
+	DeletePropagation                      metav1.DeletionPropagation      `json:"deletePropagation"`
+
+	LogRegexStr                   string            `json:"logRegex,omitempty"`
+	LogRegexesForContainersStr    map[string]string `json:"logRegexesForContainers,omitempty"`
+	SkipLogsRegexStr              string            `json:"skipLogsRegex,omitempty"`
+	SkipLogsRegexForContainersStr map[string]string `json:"skipLogsRegexForContainers,omitempty"`
 }
```
ilya-lesikov: same thing as I commented on track-readiness Operation

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+			Namespace: releaseNamespace,
+			Version:   releaseVersion,
+		},
+		instResInfos,
+		relInfos,
+		opts,
+	)
+
+	if _, err := os.Stat(path); err == nil {
+		return fmt.Errorf("plan artifact path %q already exists", path)
```
ilya-lesikov: I think I'd prefer for it to just be silently rewrited. There are probably flags for this in os.OpenFile

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+		return fmt.Errorf("stat plan artifact path %q: %w", path, err)
+	}
+
+	file, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o644)
+	if err != nil {
+		return fmt.Errorf("create plan artifact file %q: %w", path, err)
+	}
+	defer file.Close()
+
+	if err := createPlanArtifactArchive(ctx, file, artifact, secretKey, secretWorkDir); err != nil {
```
ilya-lesikov: after all the changes there is barely anything will be left in WritePlanArtifact, so just inline that createPlanArtifactArchive

pkg/action/release_plan_show.go
```diff
@@ -0,0 +1,50 @@
...
+		ShowVerboseCRDDiffs:    opts.ShowVerboseCRDDiffs,
+		ShowVerboseDiffs:       opts.ShowVerboseDiffs,
+		ShowSensitiveDiffs:     opts.ShowSensitiveDiffs,
+		ShowInsignificantDiffs: opts.ShowInsignificantDiffs,
+	}); err != nil {
+		return fmt.Errorf("log planned changes: %w", err)
+	}
+
+	return nil
+}
```
ilya-lesikov, line 59: Make a stub below for applyReleasePlanShowOptionsDefaults, please. And call it in ReleasePlanShow. Just for consistency and so that the next person changing this file will less likely reinvent this thing

Actually, some things missing. Log Level, Log Color missing, TempDirPath no defaults, maybe something else. Look at the other simple actions, please.

We should probably move a few flags to like Common group, but whatever.

pkg/action/release_plan_show.go
```diff
@@ -0,0 +1,50 @@
...
+	"os"
+
+	"github.com/samber/lo"
+
+	"github.com/werf/nelm/internal/plan"
+	"github.com/werf/nelm/pkg/log"
+)
+
+type ReleasePlanShowOptions struct {
+	plan.CalculatePlannedChangesOptions
```
ilya-lesikov: This thing is kinda internal. Either we make something like ValuesOptions out of it and place it alongside ValuesOptions and use it in the plan and whenever else, or just inline these options like we have now in the plan.

pkg/featgate/feat.go
```diff
@@ -59,6 +59,11 @@ var (
 		"Validate chart resources against specific Kubernetes resources' schemas",
 	)
 
+	FeatGatePlanFreezing = NewFeatGate(
+		"plan-freezing",
```
ilya-lesikov: You know what, can we enable it by default? I don't think it breaks anything...

cmd/nelm/release_install.go
```diff
@@ -313,6 +313,23 @@ func newReleaseInstallCommand(ctx context.Context, afterAllCommandsBuiltFuncs ma
 			return fmt.Errorf("add flag: %w", err)
 		}
 
+		if featgate.FeatGatePlanFreezing.Enabled() {
+			if err := cli.AddFlag(cmd, &cfg.PlanArtifactPath, "use-plan", "", "Use plan from specified path during release install", cli.AddFlagOptions{
```
ilya-lesikov: Please indicate somehow that this is an archive, both here and in --save-plan

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+)
+
+const PlanArtifactSchemeVersion = "v1"
+
+var ErrPlanArtifactInvalid = errors.New("plan artifact invalid")
+
+type PlanArtifact struct {
+	APIVersion string              `json:"apiVersion"`
+	Data       *PlanArtifactData   `json:"-"`
+	DataRaw    string              `json:"dataRaw,omitempty"`
```
ilya-lesikov: can we just have marshal methods in PlanArtifact for this Data instead of this Data/DataRaw manual thing?

internal/plan/planned_changes.go
```diff
@@ -32,13 +32,63 @@ type CalculatePlannedChangesOptions struct {
 
 type ResourceChange struct {
 	// Any operations on the resource after the initial one.
-	ExtraOperations []string
+	ExtraOperations []string `json:"extraOperations"`
 	// The reason for the change.
-	Reason       string
+	Reason       string `json:"reason"`
 	ResourceMeta *spec.ResourceMeta
```
ilya-lesikov: don't you need a tag here too?

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+	ReleaseLabels map[string]string
+	// ReleaseStorageDriver specifies where to store release metadata ("secrets" or "sql").
+	ReleaseStorageDriver string
+	// ReleaseStorageSQLConnection is the SQL connection string when using SQL storage driver.
+	ReleaseStorageSQLConnection string
+}
+
+type PlanArtifactData struct {
+	Changes                  []*ResourceChange          `json:"changes"`
+	DAG                      *Plan                      `json:"dag"`
```
ilya-lesikov: Let's name it Plan, like in the source code?

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+
+	if len(artifact.GetReleaseInfos()) == 0 {
+		return fmt.Errorf("%w: no release information objects found", ErrPlanArtifactInvalid)
+	}
+
+	return nil
+}
+
+func createPlanArtifactArchive(ctx context.Context, w io.Writer, artifact *PlanArtifact, secretKey, secretWorkDir string) error {
+	dataJSON, err := json.Marshal(artifact.Data) //nolint:musttag
```
ilya-lesikov: I don't think we need this //nolint:musstag anywhere. I think everything should have tags? Oh, maybe it's because of GVK?

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+
+var ErrPlanArtifactInvalid = errors.New("plan artifact invalid")
+
+type PlanArtifact struct {
+	APIVersion string              `json:"apiVersion"`
+	Data       *PlanArtifactData   `json:"-"`
+	DataRaw    string              `json:"dataRaw,omitempty"`
+	DeployType common.DeployType   `json:"deployType"`
+	Encrypted  bool                `json:"encrypted"`
+	Options    PlanArtifactOptions `json:"options"`
```
ilya-lesikov: At least Options.ReleaseStorageSQLConnection can be secret here. Hmm, maybe drop all these options into Data too? I think the user will manage without these just fine

Otherwise it's so easy for any of us to forget that you can't put anything remotely secretive there (actually, we just did :)

pkg/action/release_plan_execute.go
```diff
@@ -0,0 +1,242 @@
...
+	if prevRelease != nil {
+		newRevision = prevRelease.Version + 1
+	}
+
+	instResInfos := planArtifact.GetInstallableResourceInfos()
+	relInfos := planArtifact.GetReleaseInfos()
+
+	log.Default.Debug(ctx, "Validate plan artifact")
+
+	if err := plan.ValidatePlanArtifact(planArtifact, newRevision, opts.PlanArtifactLifetime); err != nil {
```
ilya-lesikov: btw, let's do this ValidatePlanArfifact AS EARLY as possible. Safer this way

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+	return &artifact, nil
+}
+
+func ValidatePlanArtifact(artifact *PlanArtifact, expectedRevision int, lifetime time.Duration) error {
+	if artifact == nil {
+		return fmt.Errorf("%w: plan is empty", ErrPlanArtifactInvalid)
+	}
+
+	if artifact.Timestamp.Add(lifetime).Before(time.Now().UTC()) {
+		return fmt.Errorf("%w: plan artifact is too old", ErrPlanArtifactInvalid)
```
ilya-lesikov: Please make it `plan artifact expired: was valid for <duration> until <timedate>`

cmd/nelm/release_plan_show.go
```diff
@@ -0,0 +1,124 @@
...
+
+		if err := cli.AddFlag(cmd, &cfg.TempDirPath, "temp-dir", "", "Temporary directory for operation", cli.AddFlagOptions{
+			GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
+			Group:                miscFlagGroup,
+			Type:                 cli.FlagTypeDir,
+		}); err != nil {
+			return fmt.Errorf("add flag: %w", err)
+		}
+
+		if err := cli.AddFlag(cmd, &cfg.LogColorMode, "log-color-mode", "auto", "Set log color mode. "+allowedLogColorModesHelp(), cli.AddFlagOptions{
```
ilya-lesikov: idk where it's from, but its inconsistent with other commands: "log-color-mode" instead of "color-mode", default. Grep the code for "color-mode"

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,297 @@
...
+
+		artifact.DataRaw = string(encryptedData)
+		artifact.Encrypted = true
+	} else {
+		artifact.DataRaw = string(dataJSON)
+		artifact.Encrypted = false
+	}
+
+	gzipWriter := gzip.NewWriter(w)
+	defer gzipWriter.Close()
```
ilya-lesikov: Let's please log here if it fails, not sure how harmless this function actually is.

And call it also again explicitly before "return nil" with "if err := gzipWriter.Close(); err != nil".

internal/plan/plan_artifact.go
drey: In the helm/3p-helm release.Release they use Version - https://github.com/werf/3p-helm/blob/ebc656a7ccc7c63baa1691211f3f611b0acef6bb/pkg/release/release.go#L41 Should we keep it consistent?

pkg/action/release_plan_execute.go
drey: We cannot do complete validation until we get newRevision

pkg/action/release_plan_execute.go
drey: Do we have any future plans to add ability to use plan artifact during `nelm release rollback` ? Previously, I was said, that we need something universal that can execute any plan (install or rollback).

In my opinion step 6) sounds like spaghetti code. Having plan execute separated from install makes API clearer. If started worying about duplicated code, should we also look at what happening in the `release plan install`, as it also have common parts.

internal/plan/operation.go
drey: No, it is not possible as these structs slightly different. During operation marshaling we need to extend Config to set its kind, otherwise we will not able to unmarshal Config from artifact to specific Config operation type.

pkg/action/release_plan_show.go
drey: Ok, renamed and moved to common.ResourceChangeUDiffOptions.

internal/plan/plan_artifact.go
drey: Thank you! Resolved. I assumed that it was due to GVK, but in reality it was due to tag absense for Artifact Options, and ResourceMeta attribute of ResourceChange (you identified this during review). Now golangci-lint find no issues without 'mustag' linter supression.

internal/plan/operation_config.go
drey: I double checked on this. Since golang 1.21+ it  is not required anymore. Now it is possible to marshal/unmarshal without extra wrappers.

Thank you!

internal/plan/plan_artifact.go
drey: Migrated all options to the encrypted artifact section. Thank you!

internal/plan/plan_artifact.go
drey: If we go that way, there will be too much freedom. During artifact build we still setting apiVersion, timestamps, as well as controlling which fields are mandatory. Moving this logic outside will break responsibilities.

If we bother by a lot of arguments, we can go with custom struct to pass all the required data.

What do you think?

internal/plan/plan_artifact.go
drey: I decided to go that way to protect user from artifact overwrite by occasion) Your recommendation applied.

internal/plan/plan_artifact.go
drey: Remember about encryption. 

If we wish not to have DataRaw, we have to add wrapper for marshaling/unmarshaling to PlanArtifactData. It is not enough to just add MarshalJSON/UnmarshalJSON methods to PlanArtifactData, as we need to perform encryption that requires secretKey, that we need to pass some how. JSON marshiling/unmarshaling interfaces do not allow to pass extra params. So, anyway, to perform encryption we will need either private attribute to store key, or Marshaling/Unmarshaling methods with non standard signature and an extra data type.

internal/plan/plan_artifact.go
ilya-lesikov: You are basically suggesting this style:
```go
type NewSomethingOptions {
  Arg5 string
}

// first 4 args required, the fifth is optional. The constructor basically does nothing except mapping args.
func NewSomething(arg1, arg2, arg3, arg4 string, opts NewSometingOptions) {
  return &Something{
    Arg1: arg1,
    Arg2: arg2,
    Arg3: arg3,
    Arg4: arg4,
    Arg5: opts.Arg5,
}

type Something struct {
  Arg1 string
  Arg2 string
  Arg3 string
  Arg4 string
  Arg5 string
}
```

While there are some gains (ability to separate required args from optional), the gains are marginal for an amount of boilerplate it requires (basically 3-4x of what could've been).

If this was a user-facing public API that's one thing, but this is an internal API, it's fine to have some assumptions there.

I tried to have constructors for every single struct before just for this reason (more explicit contract on what's required vs optional) and it wasn't worth it at all. So when the constructor really does nothing (like in this case) I avoid it. I'd argue this is more of a go-way to expose fields than to abstract them away with constructors and getters/setters. And the go way is basically "do it simple or you'll drown in boilerplate and maintaining this boilerplate, since there is no magic or automation for anything".

So I'd say we keep it stupid.

PS: adding even more stuff (structs) is even more boilerplate, and the issue of optional struct fields resurfaces again, just on a level higher.

pkg/action/release_plan_execute.go
ilya-lesikov: > Do we have any future plans to add ability to use plan artifact during nelm release rollback 

Yes, we do.

> Previously, I was said, that we need something universal that can execute any plan (install or rollback).

That was when I was thinking that the plan itself is enough to continue execution, whether it be install/rollback/uninstall. Turned out we also need A LOT of stuff besides the plan. This changes things, we don't (can't) have something really universal here anymore, so logic will be action-specific.

> In my opinion step 6) sounds like spaghetti code. 

Well what are other options at this point? We can't have a single Execute, there'd be three. So a lot of duplication. And a few if guards (there shouldn't be many, really) in my opinion is much simpler/clearer/maintanable.

> If started worying about duplicated code, should we also look at what happening in the release plan install, as it also have common parts.

Yea, we probably should. There is quite some duplication between actions. It's not easy to deduplicate though, because often there are small things here and there in between common blocks that make it harder to deduplicate. Also, I recently did a lot of work trying to adhere to some common flow in each action, so it can be deduplicated. I didn't get to the deduplication itself though, we'll do it later separately.

internal/plan/operation.go
ilya-lesikov: I did a bit of research and I think json.RawMessage can help us here. As far as I understand that's a conventional solution. That thing is generated by AI, but it conveys the point:

```go
package main

import (
	"encoding/json"
	"fmt"
)

type OperationType string

const (
	OpCreate OperationType = "create"
	OpUpdate OperationType = "update"
)

// OperationConfig is a marker interface (optional, but good for type safety)
type OperationConfig interface{}

type Operation struct {
	Type   OperationType   `json:"type"`
	Config OperationConfig `json:"config"`
}

type OperationConfigCreate struct {
	Resource map[string]interface{} `json:"resource"`
}

type OperationConfigUpdate struct {
	ResourceID string `json:"resource_id"` // Different fields for example
}

// Custom UnmarshalJSON
func (o *Operation) UnmarshalJSON(data []byte) error {
	// 1. Unmarshal into a temp struct to get the "type"
	type Alias Operation // Alias avoids infinite recursion
	aux := &struct {
		Config json.RawMessage `json:"config"` // Delay unmarshalling of Config
		*Alias
	}{
		Alias: (*Alias)(o),
	}

	if err := json.Unmarshal(data, &aux); err != nil {
		return err
	}

	// 2. Switch on the Type to choose the struct
	switch o.Type {
	case OpCreate:
		o.Config = &OperationConfigCreate{}
	case OpUpdate:
		o.Config = &OperationConfigUpdate{}
	default:
		return fmt.Errorf("unknown operation type: %s", o.Type)
	}

	// 3. Unmarshal the raw config into the chosen struct
	if err := json.Unmarshal(aux.Config, o.Config); err != nil {
		return err
	}

	return nil
}

func main() {
	jsonInput := `[
		{"type": "create", "config": {"resource": {"kind": "Pod"}}},
		{"type": "update", "config": {"resource_id": "12345"}}
	]`

	var ops []*Operation
	if err := json.Unmarshal([]byte(jsonInput), &ops); err != nil {
		panic(err)
	}

	for _, op := range ops {
		fmt.Printf("Type: %s, Config Type: %T, Config: %+v\n", op.Type, op.Config, op.Config)
	}
}
```

The Alias thing does looks a bit hacky, but instead of it you can keep the operationJSON struct, but then you have to duplicate all the fields in it and keep them in sync. I'd prefer the Alias approach though, less maintenance, less things to break.

You don't need here MarshalJSON method anywhere, as well as Kind() and Marshal/Unmarshal methods on the Operation. All you really need here basically is a single UnmarshalJSON() and a single switch right inside of it, nothing extra in comparison to origin/main. Should look pretty elegant (for golang anyways).

drey: Brilliant mind. Implemented. Thank you!

internal/plan/release_info.go
drey: As we do not render charts from scratch it should not affect us

internal/plan/resource_info.go
drey: Ok, will keep these fields in the artifact.

internal/plan/plan_artifact.go
drey: I made DataRaw private. Was not able to find any room for improvements.

cmd/nelm/release_install.go
```diff
@@ -325,12 +339,16 @@ func newReleaseInstallCommand(ctx context.Context, afterAllCommandsBuiltFuncs ma
 		if err := cli.AddFlag(cmd, &cfg.ReleaseNamespace, "namespace", "", "The release namespace. Resources with no namespace will be deployed here", cli.AddFlagOptions{
 			GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
 			Group:                mainFlagGroup,
-			Required:             true,
 			ShortName:            "n",
 		}); err != nil {
 			return fmt.Errorf("add flag: %w", err)
 		}
 
+		cmd.MarkFlagsOneRequired("use-plan", "release")
```
drey, line 347: I found these cobra capabilities pretty handy. I would prefer to prevent user from setting release name and namespace when it using plan, as we ignore these settings and getting it from the plan artefact.

pkg/action/release_plan_execute.go
```diff
@@ -0,0 +1,237 @@
...
+		case err := <-actionCh:
+			return err
+		case <-ctx.Done():
+			return context.Cause(ctx)
+		}
+	}
+}
+
+func releasePlanExecute(ctx context.Context, ctxCancelFn context.CancelCauseFunc, opts ReleasePlanExecuteOptions) error {
+	log.Default.Info(ctx, "Using %s plan artifact", opts.PlanArtifactPath)
```
drey: I think it makes sense to inform user that we are deploying by using plan artefact. Otherwise without debug it would be unclear from the log.

internal/plan/operation_config.go
```diff
@@ -136,8 +205,19 @@ func (c *OperationConfigTrackReadiness) IDHuman() string {
 	return c.ResourceMeta.IDHuman()
 }
 
+func (c *OperationConfigTrackReadiness) MarshalJSON() ([]byte, error) {
+	type operationConfigTrackReadinessJSON OperationConfigTrackReadiness
+
+	data, err := json.Marshal((*operationConfigTrackReadinessJSON)(c))
```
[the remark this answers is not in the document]
drey: The behaviour is different on go 1.21+ as I wrote above. https://go.dev/play/p/1KHLE1mz6pa?v=goprev

internal/plan/plan_artifact.go
drey: Thank you for the detailed explanation. I agree, if we do not suppose that some one will use it as a public API, we can simplify this part and make it more flexible. Implemented.

pkg/action/release_plan_execute.go
drey: I got your point and I refactored this part by following your recommendations. Probably, it still be handy to combine some lines into an extra function, just to minimise number of lines in the if/else block, but for now I keep it as required.

cmd/nelm/release_install.go
ilya-lesikov: You know, I think we better without this one, at least for now:
1. Inconsistent with other options.
2. Probably will be inconsistent with werf: in werf it's not going to fail on envs, but here it will.
3. Awkward for the user if the release or the namespace are part of some common envs.

internal/plan/operation_config.go
```diff
@@ -35,7 +35,7 @@ type OperationConfig interface {
 }
 
 type OperationConfigNoop struct {
-	OpID string
+	OpID string `json:"opId"`
```
ilya-lesikov: I think we should keep the abbreviations uppercase, like opID, so please restore them throughout the PR.

Let's disable the Tagliatelle linter: as far as I see there is no proper way to handle such cases, except disabling the linter altogether for such strings, which is too bothersome. It can be disabled in .golangci.yml

internal/plan/plan.go
```diff
@@ -117,6 +119,74 @@ func (p *Plan) ToDOT() ([]byte, error) {
...
+
+	sort.Slice(edges, func(i, j int) bool {
+		if edges[i].From == edges[j].From {
+			return edges[i].To < edges[j].To
+		}
+
+		return edges[i].From < edges[j].From
+	})
+
+	return json.Marshal(planJSON{ //nolint:wrapcheck
```
ilya-lesikov: prefer just wrapping the error, it's easier to just wrap instead of these nolints. Also, there can be several kinds of errors, so it's worth preserving the full error path. Naming could be weird though (marshal inside of marshal)

internal/plan/plan_artifact.go
ilya-lesikov: That's probably some old shit. Generally it's called revision, so let's use Revision. e.g. .Release.Revision and many other places

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,230 @@
...
+
+type PlanArtifact struct {
+	APIVersion string              `json:"apiVersion"`
+	Data       *PlanArtifactData   `json:"-"`
+	DeployType common.DeployType   `json:"deployType"`
+	Encrypted  bool                `json:"encrypted"`
+	Release    PlanArtifactRelease `json:"release"`
+	Timestamp  time.Time           `json:"timestamp"`
+
+	dataRaw string
```
ilya-lesikov: Well, I guess current approach is fine. Can't think of anything better. Public DataRaw with a tag could be fine too I guess.

internal/plan/plan_artifact.go
```diff
@@ -0,0 +1,230 @@
...
+
+	if artifact.Release.Namespace == "" {
+		return errors.New("release namespace is not set")
+	}
+
+	if artifact.Release.Name == "" {
+		return errors.New("release name is not set")
+	}
+
+	if artifact.Release.Version != expectedRevision {
```
ilya-lesikov: I'd say let's move this check out of this function and inline it in the action, and then call the function ValidatePlanArfifact as early as possible.

pkg/action/release_install.go
```diff
@@ -308,148 +277,181 @@ func releaseInstall(ctx context.Context, ctxCancelFn context.CancelCauseFunc, re
...
+
+		log.Default.Debug(ctx, "Build install plan")
+
+		installPlan, err = plan.BuildPlan(instResInfos, delResInfos, relInfos, plan.BuildPlanOptions{
+			NoFinalTracking: opts.NoFinalTracking,
+		})
+		if err != nil {
+			handleBuildPlanErr(ctx, installPlan, err, opts.InstallGraphPath, opts.TempDirPath, "release-install-graph.dot")
+			return fmt.Errorf("build install plan: %w", err)
+		}
```
ilya-lesikov, line 459: you know, probably leave it as is, don't bother moving this to functions for now.

We'll do the deduplication/refactoring of actions separately later.

internal/resource/resource.go
```diff
@@ -151,6 +152,35 @@ func NewInstallableResource(res *spec.ResourceSpec, releaseNamespace string, cli
 	}, nil
 }
 
+func (r *InstallableResource) MarshalJSON() ([]byte, error) {
```
ilya-lesikov: can't we get rid of it now?

internal/resource/resource.go
```diff
@@ -151,6 +152,35 @@ func NewInstallableResource(res *spec.ResourceSpec, releaseNamespace string, cli
...
+	}
+
+	type installableResourceJSON InstallableResource
+
+	if err := json.Unmarshal(data, (*installableResourceJSON)(r)); err != nil {
+		return fmt.Errorf("unmarshal installable resource: %w", err)
+	}
+
+	return nil
+}
```
ilya-lesikov: and get rid of it too?

pkg/common/options.go
```diff
@@ -217,19 +217,78 @@ func (opts *TrackingOptions) ApplyDefaults() {
...
+	// ReleaseStorageDriver specifies how release metadata is stored in Kubernetes.
+	// Valid values: "secret" (default), "configmap", "sql".
+	// Defaults to "secret" if not specified or set to "default".
+	ReleaseStorageDriver string `json:"releaseStorageDriver"`
+	// ReleaseStorageSQLConnection is the SQL connection string when using SQL storage driver.
+	// Only used when ReleaseStorageDriver is "sql".
+	ReleaseStorageSQLConnection string `json:"releaseStorageSqlConnection"`
+}
+
+type ResourceChangeUDiffOptions struct {
```
ilya-lesikov: Maybe just ResourceDiffOptions?

docs/adr/0001-nelm-plan-freezing.md
```diff
@@ -0,0 +1,211 @@
+# ADR-0001: Nelm plan freezing
```
ilya-lesikov, line 1: I'd say let's leave only Context, Problem statement and add a new section(s) about Possible solutions and Chosen solution. No implementation details, very high-level, like "Context" and "Problem statement" right now.

All in all should be like 100 lines at most ideal, 150 lines max.

cmd/nelm/release_install.go
drey: Ok, as previously namespace and release args were mandatory, I will keep this behaviour. Just made them optional if use-plan is in use.

PR #565, opened 2026-02-13
feat: deno runtime for typescript

dmmordvi:
<div id='description'>
<h3>Summary by Bito</h3>

<p>This pull request introduces Deno as a new runtime for TypeScript support in Helm charts, replacing the existing JavaScript runtime implementation based on goja and esbuild with a more secure and modern Deno-based approach that enables bundling and executing TypeScript code with controlled permissions.</p>

<details>
<ul>

<li>Introduces pkg/deno package with DenoRuntime for secure TypeScript bundling and execution in Helm charts, implementing restricted permissions for file access.</li>

<li>Adds comprehensive internal/ts package with utilities for TypeScript file handling, initialization, rendering, and bundling operations.</li>

<li>Removes dependencies on goja, goja_nodejs, esbuild, sourcemap, and related libraries, replacing them with Deno for improved security and performance.</li>

<li>Updates chart commands to include new TypeScript-specific subcommands and flags for Deno binary path configuration and bundle rebuilding.</li>

<li>Modifies chart packing and linting processes to utilize the new Deno-based bundler instead of the previous runtime.</li>

<li>Adds extensive test coverage across multiple new files for TypeScript functionality, including rendering, bundling, and package management.</li>

</ul>
</details>

</div>


cmd/nelm/chart_lint.go
```diff
@@ -261,6 +261,20 @@ func newChartLintCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*co
 			return fmt.Errorf("add flag: %w", err)
 		}
 
+		if err := cli.AddFlag(cmd, &cfg.RebuildTSBundle, "rebuild-ts", false, "Rebuild the typescript bundle even if it already exists.", cli.AddFlagOptions{
```
ilya-lesikov: Mention the limitations in the description. And maybe different flag name. How about --ignore-bundle-js? If our file is bundle.js (I don't remember) And make it cfg.IgnoreBundleJS. That we rebuild bundle under the hood if this option is true is implementation detail, the user shouldn't know it.

cmd/nelm/chart_ts_build.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+type chartTSBuildConfig struct {
+	action.ChartTSBuildOptions
+
+	LogColorMode string
+	LogLevel     string
+}
+
+func newChartTSBuildCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*cobra.Command]func(cmd *cobra.Command) error) *cobra.Command {
```
ilya-lesikov, line 23: Maybe we should name the command "ts bundle" instead of "ts build", what do you think?

cmd/nelm/chart_ts_build.go
```diff
@@ -0,0 +1,82 @@
...
+
+	afterAllCommandsBuiltFuncs[cmd] = func(cmd *cobra.Command) error {
+		if err := cli.AddFlag(cmd, &cfg.LogColorMode, "color-mode", common.DefaultLogColorMode, "Color mode for logs. "+allowedLogColorModesHelp(), cli.AddFlagOptions{
+			GetEnvVarRegexesFunc: cli.GetFlagGlobalAndLocalEnvVarRegexes,
+			Group:                miscFlagGroup,
+		}); err != nil {
+			return fmt.Errorf("add flag: %w", err)
+		}
+
+		if err := cli.AddFlag(cmd, &cfg.LogLevel, "log-level", string(log.InfoLevel), "Set log level. "+allowedLogLevelsHelp(), cli.AddFlagOptions{
```
ilya-lesikov, line 64: Check which flags are present in any command. Even if not always used. At least tempdir flag is missing.

cmd/nelm/chart_ts_init.go
```diff
@@ -13,23 +13,23 @@ import (
...
+	cfg := &chartTSInitConfig{}
 
 	cmd := cli.NewSubCommand(
 		ctx,
 		"init [PATH]",
-		"Initialize a new chart.",
-		"Initialize a new chart in the specified directory. If PATH is not specified, uses the current directory.",
-		10, // priority for ordering in help
-		chartCmdGroup,
+		"Initialize a new TypeScript chart.",
```
ilya-lesikov: It's not a TS chart initialization, but an initialization of the TS directory in the existing Helm chart

cmd/nelm/chart_ts_init.go
```diff
@@ -45,7 +45,7 @@ func newChartInitCommand(ctx context.Context, afterAllCommandsBuiltFuncs map[*co
 				cfg.ChartDirPath = args[0]
 			}
 
-			if err := action.ChartInit(ctx, cfg.ChartInitOptions); err != nil {
+			if err := action.ChartTSInit(ctx, cfg.ChartTSInitOptions); err != nil {
 				return fmt.Errorf("chart init: %w", err)
```
ilya-lesikov:
```suggestion
				return fmt.Errorf("chart ts init: %w", err)
```

cmd/nelm/groups.go
```diff
@@ -9,13 +9,15 @@ var (
 	chartCmdGroup           = cli.NewCommandGroup("chart", "Chart commands:", 90)
 	secretCmdGroup          = cli.NewCommandGroup("secret", "Secret commands:", 80)
 	dependencyCmdGroup      = cli.NewCommandGroup("dependency", "Dependency commands:", 70)
+	tsCmdGroup              = cli.NewCommandGroup("ts", "TypeScript chart commands:", 60)
```
ilya-lesikov: Let's name it just TypeScript commands. If there be commands for Helm templates they'd be Helm Templates commands, not Helm Templates Chart commands (you can have both templates and TS)

internal/chart/chart_render.go
```diff
@@ -353,10 +356,10 @@ func isLocalChart(path string) bool {
 	return filepath.IsAbs(path) || filepath.HasPrefix(path, "..") || filepath.HasPrefix(path, ".")
 }
 
-func renderJSTemplates(ctx context.Context, chartPath string, chart *helmchart.Chart, renderedValues chartutil.Values) (map[string]string, error) {
+func renderJSTemplates(ctx context.Context, chartPath string, chart *helmchart.Chart, renderedValues chartutil.Values, rebuildBundle bool, tempDirPath, denoBinaryPath string) (map[string]string, error) {
```
ilya-lesikov: inline this function

pkg/ts/init.go
```diff
@@ -65,7 +66,7 @@ func InitChartStructure(ctx context.Context, chartPath, chartName string) error
 
 	// Handle .helmignore: create or enrich
 	helmignorePath := filepath.Join(chartPath, ".helmignore")
-	if err := ensureFileEntries(helmignorePath, helmignoreContent, []string{"ts/dist/"}); err != nil {
+	if err := ensureFileEntries(helmignorePath, helmignoreContent, []string{"ts/vendor/", "ts/node_modules/"}); err != nil {
```
ilya-lesikov, line 47: do we still have vendor dir? I thought we will only have ts/bundle.js or something like that

internal/ts/init_templates.go
```diff
@@ -198,6 +240,10 @@ func chartYaml(chartName string) string {
...
 
-func packageJSON(chartName string) string {
-	return fmt.Sprintf(packageJSONTmpl, chartName, chartName)
+func denoJSON(scriptPath string) string {
+	return fmt.Sprintf(denoJSONTmpl, scriptPath)
+}
+
+func inputExample(chartName string) string {
+	return fmt.Sprintf(inputExampleContent, chartName)
 }
```
ilya-lesikov: inline all these

internal/ts/render.go
```diff
@@ -3,40 +3,50 @@ package ts
...
 
-func RenderChart(ctx context.Context, chart *helmchart.Chart, renderedValues chartutil.Values) (map[string]string, error) {
+func RenderChart(ctx context.Context, chart *helmchart.Chart, renderedValues chartutil.Values, rebuildBundle bool, chartPath, tempDirPath, denoBinaryPath string) (map[string]string, error) {
 	allRendered := make(map[string]string)
 
-	if err := renderChartRecursive(ctx, chart, renderedValues, chart.Name(), allRendered); err != nil {
-		return nil, err
+	denoRuntime := deno.NewDenoRuntime(rebuildBundle, deno.DenoRuntimeOptions{BinaryPath: denoBinaryPath})
+	if err := denoRuntime.BundleChartsRecursive(ctx, chart, chartPath); err != nil {
+		return nil, fmt.Errorf("process chart for TypeScript rendering: %w", err)
```
ilya-lesikov: unclear error message

internal/ts/render.go
```diff
@@ -3,40 +3,50 @@ package ts
...
-	if err != nil {
-		return fmt.Errorf("render files for chart %q: %w", chart.Name(), err)
-	}
+	entrypoint, bundle := deno.GetEntrypointAndBundle(chart.RuntimeFiles)
 
-	for filename, content := range rendered {
-		outputPath := path.Join(pathPrefix, filename)
-		results[outputPath] = content
-		log.Default.Debug(ctx, "Rendered output: %s", outputPath)
+	if entrypoint != "" && bundle != nil {
```
ilya-lesikov:
```suggestion
	if bundle != nil {
```

I think that's sufficient. If they really want to only pack the bundle for some reason then why not

pkg/deno/runtime.go
```diff
@@ -0,0 +1,323 @@
...
+type DenoRuntime struct {
+	binPath string
+	rebuild bool
+}
+
+func NewDenoRuntime(rebuild bool, opts DenoRuntimeOptions) *DenoRuntime {
+	return &DenoRuntime{binPath: opts.BinaryPath, rebuild: rebuild}
+}
+
+func (rt *DenoRuntime) BundleChartsRecursive(ctx context.Context, chart *helmchart.Chart, path string) error {
```
ilya-lesikov:
```suggestion
func (r *DenoRuntime) BundleChartsRecursive(ctx context.Context, chart *helmchart.Chart, path string) error {
```

Let's always have one symbol. Then every time you see single letter var you think that it might be "this"

internal/ts/render.go
```diff
@@ -58,85 +71,26 @@ func renderChartRecursive(ctx context.Context, chart *helmchart.Chart, values ch
...
+	renderDir := filepath.Join(tempDirPath, "typescript-render", chart.ChartFullPath())
+	if err := os.MkdirAll(renderDir, 0o755); err != nil {
+		return "", fmt.Errorf("create temp dir for render context: %w", err)
 	}
 
-	sourceFiles := extractSourceFiles(mergedFiles)
-	if len(sourceFiles) == 0 {
-		return map[string]string{}, nil
+	if err := writeInputRenderContext(renderedValues, chart, renderDir); err != nil {
+		return "", fmt.Errorf("build render context: %w", err)
```
ilya-lesikov:
```suggestion
		return "", fmt.Errorf("write input render context: %w", err)
```

internal/ts/render.go
```diff
@@ -3,40 +3,50 @@ package ts
...
+	if entrypoint != "" && bundle != nil {
+		content, err := renderChart(ctx, bundle, chart, values, tempDirPath, denoRuntime)
+		if err != nil {
+			return fmt.Errorf("render files for chart %q: %w", chart.Name(), err)
+		}
+
+		if content != "" {
+			outputPath := path.Join(pathPrefix, deno.ChartTSSourceDir, entrypoint)
+			results[outputPath] = content
+			log.Default.Debug(ctx, "Rendered output: %s", outputPath)
```
ilya-lesikov: Unclear debug message. Mention TS and which chart it is.

internal/ts/render.go
```diff
@@ -3,40 +3,50 @@ package ts
...
+
+	if err := renderChartRecursive(ctx, chart, renderedValues, chart.Name(), chartPath, allRendered, tempDirPath, denoRuntime); err != nil {
+		return nil, fmt.Errorf("render chart recursive: %w", err)
 	}
 
 	return allRendered, nil
 }
 
-func renderChartRecursive(ctx context.Context, chart *helmchart.Chart, values chartutil.Values, pathPrefix string, results map[string]string) error {
+func renderChartRecursive(ctx context.Context, chart *helmchart.Chart, values chartutil.Values, pathPrefix, chartPath string, results map[string]string, tempDirPath string, denoRuntime *deno.DenoRuntime) error {
```
ilya-lesikov: please refactor out results arg in this function. Just make the function return a string and handle it outside

pkg/ts/render.go
```diff
@@ -48,7 +58,10 @@ func renderChartRecursive(ctx context.Context, chart *helmchart.Chart, values ch
 			dep,
 			scopeValuesForSubchart(values, depName, dep),
 			path.Join(pathPrefix, "charts", depName),
+			filepath.Join(chartPath, "charts", depName),
 			results,
+			tempDirPath,
+			denoRuntime,
 		)
 		if err != nil {
```
ilya-lesikov, line 73: make one-line if err := func(); err != nil  statement please. Also, get rid of this useless depName var

pkg/ts/render.go
```diff
@@ -177,6 +131,36 @@ func scopeValuesForSubchart(parentValues chartutil.Values, subchartName string,
...
+	}
+
+	renderContext["Chart"] = buildChartMetadata(chart)
+
+	files := make(map[string]any, len(chart.Files))
+	for _, file := range chart.Files {
+		files[file.Name] = file.Data
+	}
+
+	renderContext["Files"] = files
```
ilya-lesikov, line 159: Is this the full context? Aren't we missing Capabilities and other things? Or maybe it's done somewhere else

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
+package deno
```
ilya-lesikov: Let's merge this package back to ts. Let's keep packages bigger, it's easier this way due to various reasons. And make everything that is not used outside of internal/ts package (i.e. not its public API) private.

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+
+	"github.com/go-resty/resty/v2"
+	"github.com/samber/lo"
+
+	"github.com/werf/3p-helm/pkg/helmpath"
+	"github.com/werf/nelm/internal/util"
+	"github.com/werf/nelm/pkg/log"
+)
+
+const version = "2.7.1"
```
ilya-lesikov: when merging this pkg back into internal/ts, make sure that symbol names are clear, e.g. here it should be denoVersion or like

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+
+var nonAlphanumRegexp = regexp.MustCompile(`[^a-zA-Z0-9]+`)
```
ilya-lesikov: don't bother, just inline

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+
+func downloadDeno(ctx context.Context, cacheDir, link string) error {
+	httpClient := util.NewRestyClient(ctx)
+	httpClient.SetTimeout(5 * time.Minute)
```
ilya-lesikov: If we do timeout here, let's make it something completely safe, like 15 mins

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+
+	log.Default.Debug(ctx, "Fetching Deno checksum from %s", checksumURL)
+
+	response, err := httpClient.R().SetContext(ctx).Get(checksumURL)
+	if err != nil {
+		return "", fmt.Errorf("download checksum from %s: %w", checksumURL, err)
+	}
+
+	if response.IsError() {
+		return "", fmt.Errorf("download checksum from %s: status code %d", checksumURL, response.StatusCode())
```
ilya-lesikov: Can we also show here the response message, not only HTTP code? It might contain important info, especially if related to 4** errors.

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	hash, _, _ := strings.Cut(strings.TrimSpace(response.String()), " ")
+	if len(hash) != 64 {
+		return "", fmt.Errorf("unexpected checksum format from %s", checksumURL)
```
ilya-lesikov: dump the result here too?

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+		return fmt.Errorf("fetch checksum: %w", err)
+	}
+
+	tmpDir, err := os.MkdirTemp(cacheDir, "download-*")
+	if err != nil {
+		return fmt.Errorf("create temp directory: %w", err)
+	}
+
+	defer func() {
+		_ = os.RemoveAll(tmpDir)
```
ilya-lesikov: let's log it to log.Default logger

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+
+	log.Default.Debug(ctx, "Downloading Deno from %s to %s", link, zipFile)
+
+	response, err := httpClient.R().SetContext(ctx).SetOutput(zipFile).Get(link)
+	if err != nil {
+		return fmt.Errorf("download Deno from %s: %w", link, err)
+	}
+
+	if response.IsError() {
+		return fmt.Errorf("download Deno from %s: status code %d", link, response.StatusCode())
```
ilya-lesikov: show response

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+func verifyChecksum(filePath, expectedHash string) error {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return fmt.Errorf("open file for checksum verification: %w", err)
+	}
+
+	defer func() {
+		_ = file.Close()
```
ilya-lesikov: log error

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+		return fmt.Errorf("verify checksum: %w", err)
+	}
+
+	reader, err := zip.OpenReader(zipFile)
+	if err != nil {
+		return fmt.Errorf("open downloaded Deno archive: %w", err)
+	}
+
+	defer func() {
+		_ = reader.Close()
```
ilya-lesikov: log error

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+	}()
+
+	binaryName := lo.Ternary(runtime.GOOS == "windows", "deno.exe", "deno")
+
+	found := false
```
ilya-lesikov: nitpick style

```suggestion
	var binaryFound bool
```

pkg/deno/runtime.go
```diff
@@ -0,0 +1,323 @@
...
+	if err := downloadDeno(ctx, cacheDir, link); err != nil {
+		return fmt.Errorf("download deno: %w", err)
+	}
+
+	rt.setBinaryPath(denoPath)
+
+	return nil
+}
+
+func (rt *DenoRuntime) getBinaryPath() string {
```
ilya-lesikov: inline, no point

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+		target = "aarch64-apple-darwin"
+	case runtime.GOOS == "windows" && runtime.GOARCH == "amd64":
+		target = "x86_64-pc-windows-msvc"
+	case runtime.GOOS == "windows" && runtime.GOARCH == "arm64":
+		target = "aarch64-pc-windows-msvc"
+	default:
+		return "", fmt.Errorf("unsupported platform: %s/%s", runtime.GOOS, runtime.GOARCH)
+	}
+
+	url := fmt.Sprintf("https://github.com/denoland/deno/releases/download/v%s/deno-%s.zip", version, target)
```
ilya-lesikov: It might not be the best idea to pull directly from github. I think there was something like rawgithub.com or whatever it was. Google the right link to download stuff like binaries from github

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+
+func getDenoFolder(downloadURL string) (string, error) {
+	hash := fmt.Sprintf("%x", sha256.Sum256([]byte(downloadURL)))
+
+	suffix := downloadURL
+	if len(suffix) > 15 {
+		suffix = suffix[len(suffix)-15:]
+	}
+
+	slug := strings.ToLower(strings.Trim(nonAlphanumRegexp.ReplaceAllString(suffix, "-"), "-"))
```
ilya-lesikov: I wonder if we have somewhere a proper slugify function for file names

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+	hash, _, _ := strings.Cut(strings.TrimSpace(response.String()), " ")
+	if len(hash) != 64 {
+		return "", fmt.Errorf("unexpected checksum format from %s", checksumURL)
+	}
+
+	return hash, nil
+}
+
+func getDenoFolder(downloadURL string) (string, error) {
+	hash := fmt.Sprintf("%x", sha256.Sum256([]byte(downloadURL)))
```
ilya-lesikov: that's way too long. Let's keep only 8 symbols out of it. Or maybe better just simpler/shorted hash, like fnv.New32a()

pkg/deno/runtime.go
```diff
@@ -0,0 +1,323 @@
...
+			_, _ = os.Stderr.Write(exitErr.Stderr)
+		}
+
+		return nil, fmt.Errorf("get deno build output: %w", err)
+	}
+
+	return output, nil
+}
+
+func (rt *DenoRuntime) setBinaryPath(path string) {
```
ilya-lesikov: pointless, get rid of this setter

pkg/deno/runtime.go
```diff
@@ -0,0 +1,323 @@
...
+func (rt *DenoRuntime) runDenoBundle(ctx context.Context, chartPath, entryPoint string) ([]uint8, error) {
+	denoBin := rt.getBinaryPath()
+	cmd := exec.CommandContext(ctx, denoBin, "bundle", entryPoint)
+	cmd.Dir = filepath.Join(chartPath, ChartTSSourceDir)
+
+	output, err := cmd.Output()
+	if err != nil {
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			_, _ = os.Stderr.Write(exitErr.Stderr)
```
ilya-lesikov: shouldn't be dumping it to stderr without passing it through the log.Default() logger. Maybe log.Default().Error

pkg/deno/runtime.go
```diff
@@ -0,0 +1,323 @@
...
+		bundleRes, err := rt.runDenoBundle(ctx, path, entrypoint)
+		if err != nil {
+			return fmt.Errorf("build TypeScript bundle: %w", err)
+		}
+
+		if rt.rebuild && bundle != nil {
+			chart.RemoveRuntimeFile(ChartTSBundleFile)
+		}
+
+		chart.AddRuntimeFile(ChartTSBundleFile, bundleRes)
```
ilya-lesikov: What will happen when rt.rebuild == false && bundle != nil? Do we get two bundles (old and new) in RuntimeFiles?

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+func unzipBinary(cacheDir string, file *zip.File) error {
+	destPath := filepath.Join(cacheDir, filepath.Base(file.Name))
+
+	denoFile, err := os.Create(destPath)
+	if err != nil {
+		return fmt.Errorf("create file for Deno binary: %w", err)
+	}
+
+	defer func() {
+		_ = denoFile.Close()
```
ilya-lesikov: log error

pkg/deno/downloader.go
```diff
@@ -0,0 +1,226 @@
...
+	fileReader, err := file.Open()
+	if err != nil {
+		return fmt.Errorf("open file %s in Deno archive: %w", file.Name, err)
+	}
+
+	defer func() {
+		_ = fileReader.Close()
+	}()
+
+	limitReader := io.LimitReader(fileReader, 200*1024*1024)
```
ilya-lesikov: I don't think we need this limiter here. Let's not overcomplicate. If its more than 200mb for whatever reason why should we care really

pkg/deno/runtime.go
```diff
@@ -0,0 +1,323 @@
...
+	ChartTSEntryPointJS = "src/index.js"
+	// ChartTSEntryPointTS is the TypeScript entry point path.
+	ChartTSEntryPointTS = "src/index.ts"
+	// ChartTSSourceDir is the directory containing TypeScript sources in a Helm chart.
+	ChartTSSourceDir = "ts/"
+	// RenderInputFileName is the name of the input file with context for the Deno app.
+	RenderInputFileName = "input.yaml"
+	// RenderOutputFileName is the name of the output file with rendered manifests from the Deno app.
+	RenderOutputFileName = "output.yaml"
+)
```
ilya-lesikov: I'd say let's move these constants into pkg/common, they might be useful in different contexts without any logic of this package.

pkg/deno/runtime.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func (rt *DenoRuntime) BundleChartsRecursive(ctx context.Context, chart *helmchart.Chart, path string) error {
+	entrypoint, bundle := GetEntrypointAndBundle(chart.RuntimeFiles)
+	if entrypoint == "" {
+		return nil
+	}
+
+	if bundle == nil || rt.rebuild {
+		if err := rt.ensureBinary(ctx); err != nil {
```
ilya-lesikov: we are ensuring binary every time we recurse into a chart, but really only need to do it once. Maybe we better do downloading/caching outside of denoruntime? All we need is to go through Chart{}'s recursively and check if they have ts dir? And if there is one then download the thing to cache (unless custom path is specified)

Frankly, I don't think we need DenoRuntime class in here. Just make a few functions out of it, one will download binary, the other one will build bundle.js-es, the third one will run the bundles. There is no state to share so we don't need any classes for now

ilya-lesikov (changes requested):

pkg/ts/init.go
dmmordvi: In theory, user can change `deno.json` and set [vendoring](https://docs.deno.com/runtime/fundamentals/modules/#vendoring-remote-modules), so it is better to ignore this dir

pkg/deno/runtime.go
dmmordvi: I used DenoRuntime class only because of 3p-helm... It is simpler to pass it like class and run BundleChartsRecursive in chart pack, and everything will work under the hood (check ts dir, download deno, bundle and add files into RuntimeFiles)

cmd/nelm/chart_ts_build.go
dmmordvi: I am not sure, but as i remember we decided to use `ts build`, because we already have `werf bundle` commands

pkg/ts/render.go
dmmordvi: Other stuff comes from [here](https://github.com/werf/nelm/blob/e17d51190720367f43cd3841720500091a71110c/internal/ts/render.go#L135):
`renderContext := renderedValues.AsMap()`

pkg/deno/runtime.go
dmmordvi: Theoretically it shouldn't because of this condition https://github.com/werf/nelm/blob/30423bf57dd25328b5af5eb9ae90e8687b1c9206/pkg/deno/runtime.go#L59.
But i will simplify condition to `if bundle != nil {`

pkg/deno/downloader.go
dmmordvi: Decided to keep package

pkg/deno/downloader.go
dmmordvi: Decided to keep current links

pkg/deno/downloader.go
dmmordvi: Checked - there is no ready slugify in nelm or common-go/3p-helm. there is only `pkg/slug` in werf...

cmd/nelm/chart_ts_init.go
dmmordvi: Actually, we ensure chart structure too: we create Chart.yaml and values.yaml if it doesn't exist + we check .helmignore/.gitignore have necessary keys. Should we mention in cmd description that we ensure chart structure too?

pkg/deno/downloader.go
dmmordvi: Added gosimple/slug

cmd/nelm/chart_ts_init.go
ilya-lesikov: Chart.yaml shouldn't be created. Alright, let's say it's not "an initialization of the TS directory", but "an initialization of files needed to render manifests via TS" or something. Write whatever, as long as the user don't think we would generate the chart from scratch.

pkg/deno/downloader.go
ilya-lesikov: Alright

pkg/deno/runtime.go
ilya-lesikov: Oh, alright, I see now

cmd/nelm/common.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/werf/nelm/pkg/log"
 )
 
+const IgnoreBundleJSFlagDescription = "Do not use the existing bundle.js file. Requires TypeScript source files and Deno to rebuild."
```
ilya-lesikov: let's just inline it, we'll solve this duplication later, all at once

ilya-lesikov (approved):

PR #586, opened 2026-03-26
feat: expose plan artifact in actions api

drey:


pkg/action/release_plan_install.go
```diff
@@ -112,8 +113,47 @@ type ReleasePlanInstallOptions struct {
 	Timeout time.Duration
 }
 
+func LogPlannedChanges(ctx context.Context, planArtifact *planartifact.Artifact, opts common.ResourceDiffOptions) error {
```
ilya-lesikov: Let's not expose LogPlannedChanges. Instead pass the opts.LegacyPlanArtifact into ReleasePlanShow as we did in ReleaseInstall.

pkg/plan/artifact.go
```diff
@@ -13,39 +13,40 @@ import (
...
+const ArtifactSchemeVersion = "v1"
 
-type PlanArtifact struct {
-	APIVersion string              `json:"apiVersion"`
-	Data       *PlanArtifactData   `json:"-"`
-	DataRaw    string              `json:"dataRaw"`
-	DeployType common.DeployType   `json:"deployType"`
-	Encrypted  bool                `json:"encrypted"`
-	Release    PlanArtifactRelease `json:"release"`
+type Artifact struct {
```
ilya-lesikov: Hmm, it either belongs to pkg/actions or pkg/plan, doesnt worth a separate package. And I'm thinking we better leave it at pkg/actions, because many things here are way too high-level, orchestration-level, like reading/writing plan to the disk.

ilya-lesikov (changes requested):

PR #593, opened 2026-04-10
feat: ability to init ts files with WerfRenderContext type

dmmordvi:


pkg/ts/init.go
```diff
@@ -11,7 +11,14 @@ import (
 	"github.com/werf/nelm/pkg/log"
 )
 
-const denoBuildScript = "deno bundle --output=dist/bundle.js src/index.ts"
+const (
+	defaultRenderContextType = "RenderContext"
```
ilya-lesikov: Move it where WerfRenderContextType is defined and name it something like TSGenericRenderContextType and TSWerfRenderContextType. Then add a TSDefaultRenderContextType = TSGenericRenderContextType

pkg/ts/init_templates.go
```diff
@@ -6,14 +6,14 @@ const (
     "build": "%s"
   },
   "imports": {
-    "@nelm/chart-ts-sdk": "npm:@nelm/chart-ts-sdk@^0.1.3"
+    "@nelm/chart-ts-sdk": "npm:@nelm/chart-ts-sdk@^0.1.4"
   }
 }
 `
-	deploymentTSContent = `import type { RenderContext } from '@nelm/chart-ts-sdk';
+	deploymentTSTmpl = `import type { {{RenderContextType}} } from '@nelm/chart-ts-sdk';
```
ilya-lesikov: well, since we started templating it, then let's do it properly via go text/template engine. Just init a new blank engine and pass a context ($) into it with all you need. Will be easier to extend later. And the template will become this instead: `{{ $.RenderContextType }}`

ilya-lesikov (changes requested):

PR #602, opened 2026-04-24
feat: forbid resource recreation if helm.sh/resource-policy annotatio…

drey:
…n set

### DIFF
--- a/pkg/plan/plan_build.go
+++ b/pkg/plan/plan_build.go
@@ -8,7 +8,6 @@ import (
 	"github.com/samber/lo"
 
 	"github.com/werf/nelm/pkg/common"
-	helmkube "github.com/werf/nelm/pkg/helm/pkg/kube"
 	helmreleasecommon "github.com/werf/nelm/pkg/helm/pkg/release/common"
 	helmrelease "github.com/werf/nelm/pkg/helm/pkg/release/v1"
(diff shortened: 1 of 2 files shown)


pkg/plan/plan_build.go
```diff
@@ -425,6 +426,9 @@ func addInstallResourceOps(plan *Plan, infos []*InstallableResourceInfo) error {
 			}
 			chain.AddOperation(createOp).Stage(stg)
 		case ResourceInstallTypeRecreate:
+			if _, found := info.LocalResource.Annotations[helmkube.ResourcePolicyAnno]; found {
```
ilya-lesikov: Please use AnnotationKeyHumanResourcePolicy

pkg/plan/plan_build.go
```diff
@@ -425,6 +426,9 @@ func addInstallResourceOps(plan *Plan, infos []*InstallableResourceInfo) error {
...
+				return fmt.Errorf("%s re-creation is prohibited due to %q annotation set", info.IDHuman(), helmkube.ResourcePolicyAnno)
```
ilya-lesikov: I think we can do this check earlier, on ResourceInfo level. And I guess just say something generic, like deletions are prohibited for this resource and this is why we won't recreate it. Or we'll add some another mechanism that can forbid deletions and will forget about this error message

PR #612, opened 2026-04-29
refactor: join Hooks and Resources as ResourceSpecs

dmmordvi:


ilya-lesikov (approved):

pkg/action/release_get.go
```diff
@@ -60,14 +60,12 @@ type ReleaseGetOptions struct {
 }
 
 type ReleaseGetResultV1 struct {
```
ilya-lesikov: change to V2, and APIVersion too

PR #614, opened 2026-05-07
feat: fail on invalid annotations or labels

drey:


pkg/resource/spec/resource_spec.go
```diff
@@ -57,7 +57,22 @@ func NewResourceSpecFromManifest(manifest, releaseNamespace string, opts Resourc
 		return nil, fmt.Errorf("decode resource (file: %q): %w", opts.FilePath, err)
 	}
 
-	return NewResourceSpec(obj.(*unstructured.Unstructured), releaseNamespace, opts), nil
+	unstruct := obj.(*unstructured.Unstructured)
+
+	if !opts.DropInvalidAnnotationsAndLabels {
```
ilya-lesikov: nitpick: `if opts.DropInvalidAnnotationsAndLabels` would be easier to understand, otherwise the `else` clause is effectively a double negation

pkg/resource/spec/util.go
```diff
@@ -152,3 +155,25 @@ func setAnnotationsAndLabels(res *unstructured.Unstructured, annotations, labels
 		res.SetLabels(lbls)
 	}
 }
+
+func stripInvalidEntries(filePath string, obj map[string]interface{}, fields ...string) map[string]string {
+	result := make(map[string]string)
+
+	ctx := logboek.NewContext(context.Background(), logboek.DefaultLogger())
```
ilya-lesikov: just pass ctx everywhere, it's fine, we'll need it sooner or later anyways

ilya-lesikov (approved):

PR #615, opened 2026-05-12
feat: improve runtime annotations and labels handling

drey:


pkg/action/chart_render.go
```diff
@@ -282,6 +282,7 @@ func ChartRender(ctx context.Context, opts ChartRenderOptions) (*ChartRenderResu
 
 	releasableResSpecs, err := spec.BuildReleasableResourceSpecs(ctx, opts.ReleaseNamespace, transformedResSpecs, []spec.ResourcePatcher{
 		spec.NewExtraMetadataPatcher(opts.ExtraAnnotations, opts.ExtraLabels),
+		spec.NewExtraMetadataPatcher(opts.ExtraRuntimeAnnotations, opts.ExtraRuntimeLabels),
```
ilya-lesikov, line 285: This is definitely a hack. This makes resources in Release{} in ChartRender() different from ReleaseInstall(), I don't like it. It works now, but it is hacky and counterintuitive

Any other ways? We can change the whole flow if needed, maybe make it closer to what was done in ReleaseInstall

pkg/plan/resource_info.go
```diff
@@ -816,15 +819,19 @@ func resourceInstallType(ctx context.Context, localRes *resource.InstallableReso
...
 
 	diffableGetObj := spec.CleanUnstruct(getObj, spec.CleanUnstructOptions{
-		CleanHelmShAnnos: true,
-		CleanWerfIoAnnos: true,
-		CleanRuntimeData: true,
+		CleanHelmShAnnos:   true,
+		CleanWerfIoAnnos:   true,
+		CleanRuntimeData:   true,
+		ExcludeAnnotations: extraRuntimeAnnotations,
+		ExcludeLabels:      extraRuntimeLabels,
```
ilya-lesikov: Remind me, what are we trying to do here? Now it seems like werf.io/version or werf.io/release-channel will start triggering resource updates, which we don't want.

Also, here'd be better to have just a []string instead of map[string]string for these Exclude* params

drey: `ExcludeAnnotations` and `ExcludeLabels` are populated by `--runtime-annotations` and `--runtime-labels` flags respectively. These constructs do not change other annotations/labels behaviour. These params required to skip resource update if these annotachions changes as described in respective cli arguments:

```
      --runtime-annotations={}
          Add annotations which will not trigger resource updates to all resources. Vars: $NELM_RUNTIME_ANNOTATIONS_*,
          $NELM_CHART_LINT_RUNTIME_ANNOTATIONS_*
      --runtime-labels={}
          Add labels which will not trigger resource updates to all resources. Vars: $NELM_RUNTIME_LABELS_*,
          $NELM_CHART_LINT_RUNTIME_LABELS_*
```

It make sense to store data as a map, to be able to perform check with O(1)` time complexity.

pkg/action/chart_render.go
drey: Refactored as discussed on the call. Thanks!

pkg/plan/resource_info.go
ilya-lesikov: IMO not worth it. No gain for real-life scenarios, but makes code/API ambiguous/unclear.

ilya-lesikov (approved):

PR #617, opened 2026-05-12
feat: support external dependencies for deploy-dependency and delete-dependency annotations

dmmordvi:


pkg/resource/dependency.go
```diff
@@ -10,12 +10,17 @@ import (
 )
 
 // Represents a dependency on a Kubernetes resource in the Helm release.
-type InternalDependency struct {
+type Dependency struct {
 	*spec.ResourceMatcher `json:"resourceMatcher"`
 
+	ResourceMeta  *spec.ResourceMeta   `json:"resourceMeta"`
```
ilya-lesikov: You have both ResourceMatcher and ResourceMeta, bad/unclear design

pkg/resource/metadata.go
```diff
@@ -520,19 +521,69 @@ func manualInternalDeleteDependencies(meta *spec.ResourceMeta) []*InternalDepend
...
+				if err != nil {
+					return nil, fmt.Errorf("invalid value for \"external\" property of dependency %q: %w", depID, err)
+				}
+
+				dep.External = val
+			}
+
+			if dep.External {
+				dep.MinMatches = 1
+				dep.MaxMatches = 30
```
ilya-lesikov: move these as constants to common pkg

pkg/resource/metadata.go
```diff
@@ -583,11 +634,61 @@ func manualInternalDeployDependencies(meta *spec.ResourceMeta) []*InternalDepend
...
```
ilya-lesikov: to constants

pkg/resource/metadata.go
```diff
@@ -773,6 +874,19 @@ func validateDeleteDependencies(meta *spec.ResourceMeta) error {
 					default:
 						panic(fmt.Sprintf("unexpected type %T for property %q", pv, propKey))
 					}
+				case "external":
+					switch pv := propVal.(type) {
+					case string:
+						switch pv {
+						case "auto", "true", "false":
```
ilya-lesikov: to constants to common pkg, and use them everywhere in this PR

pkg/resource/resource.go
```diff
@@ -108,7 +108,27 @@ func NewInstallableResource(res *spec.ResourceSpec, releaseNamespace string, cli
...
+	internalDeps := lo.Filter(manDeps, func(item *Dependency, _ int) bool {
+		return !item.External
+	})
+
+	if err := validateDeleteDependencies(res.ResourceMeta); err != nil {
+		return nil, fmt.Errorf("validate delete dependencies: %w", err)
+	}
+
+	// Catches validation error for External delete dependencies
+	if _, err := manualDeleteDependencies(res.ResourceMeta, otherResourceMetaList, releaseNamespace); err != nil {
```
ilya-lesikov: if we do separately validateDeploy(Delete)Dependencies, can't we write manualDeploy(Delete)Dependencies in a way that they rely on the previous call of validate... functions? So that inside manualDeploy(Delete)Dependencies we often won't check errors (panic instead), etc, so no error will be returned at all? Look at how we do other functions like this where validate is separate

pkg/common/common.go
```diff
@@ -96,17 +96,23 @@ const (
...
+	DefaultLogColorMode                 = log.LogColorModeAuto
+	DefaultNetworkParallelism           = 30
+	DefaultProgressPrintInterval        = 5 * time.Second
+	DefaultQPSLimit                     = 30
+	DefaultReleaseHistoryLimit          = 10
 	// DefaultResourceValidationKubeVersion Kubernetes version to use during resource validation by kubeconform
 	DefaultResourceValidationKubeVersion = "1.35.0"
+	DependencyExternalAuto               = "auto"
+	DependencyExternalFalse              = "false"
+	DependencyExternalTrue               = "true"
```
ilya-lesikov: Make a type DependencyExternal string for those

pkg/resource/resource.go
```diff
@@ -209,8 +200,8 @@ func NewDeletableResource(resourceSpec *spec.ResourceSpec, otherResourceSpecs []
 	})
 
 	var manDeps []*Dependency
-	if err := validateDeleteDependencies(resourceSpec.ResourceMeta); err == nil {
-		manDeps, _ = manualDeleteDependencies(resourceSpec.ResourceMeta, otherResourceMetaList, releaseNamespace)
+	if err := validateDeleteDependencies(resourceSpec.ResourceMeta, otherResourceMetaList); err == nil {
```
ilya-lesikov: Huh? We should fail here if err != nil

ilya-lesikov (approved):

PR #625, opened 2026-05-18
feat: remove deprecated external dependencies annotations

dmmordvi:


pkg/resource/metadata.go
```diff
@@ -852,6 +722,18 @@ func ownership(meta *spec.ResourceMeta, releaseNamespace string, storeAs common.
 	}
 }
 
+func rejectDeprecatedExternalDependencies(meta *spec.ResourceMeta) error {
+	deprecatedAnnotationPattern := regexp.MustCompile(`^(?P<id>.+)\.external-dependency\.werf\.io(/resource|/namespace)?$`)
+
+	if annotations, found := spec.FindAnnotationsOrLabelsByKeyPattern(meta.Annotations, deprecatedAnnotationPattern); found {
+		for key := range annotations {
+			return fmt.Errorf("annotation %q in resource %q is no longer supported, use %q or %q instead", key, meta.IDHuman(), common.AnnotationKeyHumanDeployDependency, common.AnnotationKeyHumanDeleteDependency)
```
ilya-lesikov: Let's do a warning instead. Maybe the external deps they defined in the charts aren't that important or they are already satisfied after the first release. Deleting this annotation is not a good change from the UX perspective and is more because of internal reasons, so let's not scream about it.

ilya-lesikov (approved):

PR #626, opened 2026-05-21
feat: case insensitive group/version/kind for `deploy/delete-dependency`

dmmordvi:


pkg/plan/operation_config.go
```diff
@@ -112,7 +112,8 @@ func (c *OperationConfigDelete) IDHuman() string {
 }
 
 type OperationConfigTrackReadiness struct {
-	ResourceMeta *spec.ResourceMeta `json:"resourceMeta"`
+	ResourceMeta               *spec.ResourceMeta `json:"resourceMeta"`
+	CaseInsensitiveGVKMatching bool               `json:"caseInsensitiveGVKMatching"`
```
ilya-lesikov: why do we need this? Is there any case? If that's the only behavior, let's just force it and no option

PR #630, opened 2026-06-03
fix: deploy ns on pre-pre-install stage

drey:


pkg/resource/spec/util.go
```diff
@@ -84,6 +84,10 @@ func IsCRD(groupKind schema.GroupKind) bool {
 	}
 }
 
+func IsNamespace(gvk schema.GroupVersionKind) bool {
+	return gvk.Group == "" && gvk.Version == "" && gvk.Kind == "Namespace"
```
ilya-lesikov: let's get rid of version check here, only group

pkg/resource/metadata.go
ilya-lesikov: hook-namespace (for whatever reason) should work similarly to CRD

ilya-lesikov: should work similarly to CRD

ilya-lesikov (approved):

pkg/resource/metadata.go
```diff
@@ -1462,6 +1470,10 @@ func weight(meta *spec.ResourceMeta, hasManualInternalDeps bool) *int {
 		return nil
 	}
 
+	if spec.IsNamespace(meta.GroupVersionKind) {
+		return lo.ToPtr(-99)
```
ilya-lesikov: just make 0 weight, as with CRD. This is a special stage, nothing except CRD/ns can be here now

ilya-lesikov (approved):

PR #631, opened 2026-06-04
feat: add automatic deploy dependency detection for KEDA ScaledObjects

### LINKED ISSUE #456: KEDA ScaledObject is trying to be deployed before deployment
### Before proceeding

- [x] I didn't find a similar [issue](https://github.com/werf/nelm/issues)

### Version

1.16.0

### How to reproduce

1. use keda autoscaler in chart
2. try to deply
3. get error

### Result

Unexpected error when reading response body: context canceled Error: install: failed release "r" (namespace: "ns"): 2 errors occurred: * execute release install plan: wait for operations completion: execute operation: create resource: server-side apply resource "ScaledObject/app-scaledobject": admission webhook "vscaledobject.kb.io" denied the request: Deployment.apps "app" not found

### Expected result

Everything is deployed properly.

### Additional information

_No response_

Sharvash:
nelm now automatically deploys a KEDA `ScaledObject` after the workload referenced by its `spec.scaleTargetRef`, so the ordering no longer has to be declared by hand.

Previously, when a `ScaledObject` and its target workload (e.g. a `Deployment`) belonged to the same release, nelm put both in the same deploy stage with no ordering between them. If the `ScaledObject` reached the API server before its target existed, the install broke in one of two ways:
- with a `cpu`/`memory` trigger, KEDA's validating webhook resolves the target and denies the request (the error reported in #456): `admission webhook "vscaledobject.kb.io" denied the request: Deployment.apps "app" not found`
- with any other trigger the object is admitted, but the KEDA controller cannot reconcile it and it stays `Ready=False` with `reason=ScaledObjectCheckFailed`

The only way around it was a manual dependency annotation:
```yaml
werf.io/deploy-dependency-app: state=present,kind=Deployment,name=app
```

`internalDeployDependencies` already derives implicit dependencies for built-in workloads, RBAC bindings and others. This adds a `keda.sh/ScaledObject` case that reads `spec.scaleTargetRef` and emits a `present` dependency on the referenced workload:
- `name` - required; if it is missing the `ScaledObject` is left untouched
- `kind` - defaults to `Deployment`
- `apiVersion` - defaults to `apps/v1`, and the group is parsed from it

The dependency is `present`, not `ready`: nelm waits only for the target to exist, which is all KEDA needs, instead of blocking on its readiness. `ScaledJob` is intentionally left out — it uses `jobTargetRef` and creates its own Jobs, so it has no such dependency.

Verified against a live cluster running KEDA: with the change the `ScaledObject` is always created after its target and reconciles cleanly, and the #456 webhook denial reproduces without it.

Closes #456


ilya-lesikov:
Hi, please, resolve conflicts, but don't force-push anymore, otherwise I need to look at all the commits again.

ilya-lesikov:
Thanks!

PR #635, opened 2026-06-10
feat: native Chart v3 and Release v2 support

drey:


pkg/action/chart_ts_build.go
```diff
@@ -58,16 +60,26 @@ func ChartTSBuild(ctx context.Context, opts ChartTSBuildOptions) error {
...
+		return fmt.Errorf("create chart accessor: %w", err)
 	}
 
-	if err = ts.BundleChartsRecursive(ctx, chart, absPath, true, opts.DenoBinaryPath); err != nil {
+	if err = ts.BundleChartsRecursive(ctx, chartAccessor, absPath, true, opts.DenoBinaryPath); err != nil {
 		return fmt.Errorf("process chart: %w", err)
 	}
 
-	bundles := lo.Filter(chart.Raw, func(file *chartcommon.File, _ int) bool {
+	var rawFiles []*chartcommon.File
```
ilya-lesikov: won't it be easier to expose these raw files through a method of an interface?

pkg/release/history.go
```diff
@@ -189,3 +192,18 @@ func BuildHistory(releaseName string, historyStorage ReleaseStorager, opts Histo
 		opts,
 	), nil
 }
+
+func touchReleaseDeployTimes(releaser helmrel.Releaser, t time.Time) error {
+	switch r := releaser.(type) {
+	case *helmrelease.Release:
```
ilya-lesikov: make a method to configure this in Releaser, then get rid of this touchReleaseDeployTimes function

pkg/release/history.go
```diff
@@ -67,11 +71,11 @@ func (h *History) DeleteRelease(ctx context.Context, name string, revision int)
 
 	rel, err := h.storage.Delete(name, revision)
 	if err != nil {
-		return fmt.Errorf("uninstall release %q (namespace: %q, revision: %q): %w", rel.Name, rel.Namespace, rel.Version, err)
+		return fmt.Errorf("uninstall release %q (revision: %d): %w", name, revision, err)
```
ilya-lesikov, line 72: return namespace back

pkg/release/release.go
```diff
@@ -185,28 +200,59 @@ func NewRelease(name, namespace string, revision int, deployType common.DeployTy
...
+			Name: name,
+			Info: &v2release.Info{
+				Status: status,
+				Notes:  opts.Notes,
+			},
+			Chart:    chartObj,
+			Config:   releaseConfig,
+			Manifest: strings.Join(regularResources, "\n---\n"),
+			Hooks: lo.Map(hookResources, func(h *helmrelease.Hook, _ int) *v2release.Hook {
+				return v1HookToV2Hook(h)
```
ilya-lesikov: Not how it should be done. We have a helper  releaseutil.HookManifestToHook() for v1 hooks, we need a similar for v2 hooks. And higher, where we call this helper, we should call one or another dependening on chart version. So that here we don't need to convert v1 to v2 hook and this v1HookToV2Hook can be deleted.

pkg/release/release_storage.go
```diff
@@ -84,6 +104,24 @@ type ReleaseStorageOptions struct {
...
 }
 
+func CopyReleaserWithStatus(releaser helmrel.Releaser, status helmreleasecommon.Status) (helmrel.Releaser, error) {
+	copied, err := copystructure.Copy(releaser)
+	if err != nil {
+		return nil, fmt.Errorf("deep copy release: %w", err)
+	}
+
+	switch r := copied.(type) {
+	case *helmrelease.Release:
```
ilya-lesikov: Make a setter for this in Accessor and use Accessor. Then get rid of this CopyReleaserWithStatus helper

pkg/release/release_storage.go
```diff
@@ -123,3 +161,46 @@ func NewReleaseStorage(ctx context.Context, namespace, storageDriver string, cli
 
 	return &storageAdapter{storage: storage}, nil
 }
+
+func ReleaserToV1Release(releaser helmrel.Releaser) (*helmrelease.Release, error) {
```
ilya-lesikov, line 147: mark callings of this function with something like # XXX: must convert to v1 for now, since support of v2 releases in storage drivers doesn't yet exist

pkg/release/stored_release.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	v2release "github.com/werf/nelm/pkg/helm/intern/release/v2"
+	helmrel "github.com/werf/nelm/pkg/helm/pkg/release"
+	helmrelease "github.com/werf/nelm/pkg/helm/pkg/release/v1"
+)
+
+type StoredRelease struct {
```
ilya-lesikov: hmm, do we really need this struct? Can't we like add these methods to Accessor or something, so it can be marshaled to/from both v1 and v2 releases?

pkg/plan/plan_execute.go
```diff
@@ -325,7 +326,12 @@ func execOpCreate(ctx context.Context, op *Operation, releaseNamespace string, c
 func execOpCreateRelease(ctx context.Context, op *Operation, history release.Historier) error {
 	opConfig := op.Config.(*OperationConfigCreateRelease)
 
-	if err := history.CreateRelease(ctx, opConfig.Release); err != nil {
+	rel, err := helmrel.NewAccessor(opConfig.Release.Releaser)
```
ilya-lesikov: do we ever need to operate with just a Releaser? Sounds useless. It feels like all through nelm codebase (except maybe in pkg/helm) we need to only operate with Accessor. It has Releaser inside if anything.

ilya-lesikov (changes requested):

PR #645, opened 2026-06-22
feat: explain required install with no planned resource changes

### LINKED ISSUE #512: `plan` doesn't show exact reason for new release, if no resource changes
### Before proceeding

- [x] I didn't find a similar [issue](https://github.com/werf/werf/issues)

### Problem

WERF_EXIT_CODE: "true"

We use werf plan to obtain information about what will change during werf converge, but if the previous rollout was unsuccessful but all changes were applied, werf plan issues a warning but also writes that there is nothing to change unnecessarily. 

Image

### Solution (if you have one)

If diff is equal to 0, output raw data from nelm about the reason for the changes.

### Additional information

Version: v2.55.5

Sharvash:
nelm's `release plan install` command now explains why a release must still be installed even when no Kubernetes resource changes are planned.

Previously, when the plan found no resource changes but the release still had to be installed, it printed only a generic yellow warning:
```
No resource changes planned, but still must install release "app" (namespace: "ns")
```
There was no way to tell what triggered the install — a failed previous release, changed values, updated manifests, or something else.

`IsReleaseUpToDate` is the single place that decides whether a release is up to date. It now returns a typed `ReleaseOutdatedReason` next to its boolean result instead of just `(bool, error)`, and the check is split into separate guard clauses that each return their specific reason:
- there is no previously deployed release
- the previously deployed release was not successful
- the release notes changed
- the release values changed
- the release hooks changed
- the release manifests changed

`release plan install` appends the reason to the existing warning through a small `releaseMustInstallMessage` helper, so it now reads:
```
No resource changes planned, but still must install release "app" (namespace: "ns") because the release values changed
```
The helper leaves the message untouched when the reason is empty, so there is never a dangling " because". The `release install` and `release rollback` commands ignore the reason and keep their current "already as desired" messages unchanged.

Table-driven unit tests cover every branch of `IsReleaseUpToDate` and the message-builder guard. Build, lint and unit tests are green.

Closes #512

### DIFF
--- a/pkg/release/release.go
+++ b/pkg/release/release.go
@@ -22,6 +22,18 @@ import (
 	"github.com/werf/nelm/pkg/resource/spec"
 )
 
+const (
+	ReleaseOutdatedReasonNone                     ReleaseOutdatedReason = ""
+	ReleaseOutdatedReasonNoPreviousRelease        ReleaseOutdatedReason = "there is no previously deployed release"
+	ReleaseOutdatedReasonReleaseStatusNotDeployed ReleaseOutdatedReason = "the previously deployed release was not successful"
+	ReleaseOutdatedReasonNotesChanged             ReleaseOutdatedReason = "the release notes changed"
+	ReleaseOutdatedReasonValuesChanged            ReleaseOutdatedReason = "the release values changed"
+	ReleaseOutdatedReasonHooksChanged             ReleaseOutdatedReason = "the release hooks changed"
+	ReleaseOutdatedReasonManifestsChanged         ReleaseOutdatedReason = "the release manifests changed"
+)
+
+type ReleaseOutdatedReason string
+
 type ReleaseOptions struct {
 	InfoAnnotations map[string]string
 	Labels          map[string]string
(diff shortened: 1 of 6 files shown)


pkg/release/release.go
```diff
@@ -30,51 +42,57 @@ type ReleaseOptions struct {
 
 // Check if the new Release is up-to-date compared to the old Release. It doesn't check any
 // resources of the release in the cluster, just compares Release objects.
-func IsReleaseUpToDate(oldRel, newRel *helmrelease.Release) (bool, error) {
+func IsReleaseUpToDate(oldRel, newRel *helmrelease.Release) (bool, ReleaseOutdatedReason, error) {
```
ilya-lesikov: I like the idea generally, but I wonder if we can make it cleaner. Later I'd like to be able to also show (at least in some cases) diffs on what changed, e.g. which values changed exactly or in what way manifests changed.

Maybe we shouldn't return isUpToDate + reason, but should return IsReleaseUpToDateResult instead? It can encapsulate both, can also contain additional info for displaying stuff later.

ilya-lesikov (changes requested):

pkg/release/release.go
Sharvash: Done, `IsReleaseUpToDate` now returns an `IsReleaseUpToDateResult` that carries `UpToDate` and `Reason`, instead of the separate bool and reason. That gives us a place to add the diff details later (which values changed, how the manifests differ) without touching the signature again.

ilya-lesikov (approved):

PR #657, opened 2026-07-13
feat: add --lookup-resources flag for offline lookup in render/lint commands

dmmordvi:
<a href="https://cubic.dev/pr/werf/nelm/pull/657?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/action/chart_lint.go
+++ b/pkg/action/chart_lint.go
@@ -98,7 +98,9 @@ type ChartLintOptions struct {
 	// template function resolves against in local mode (Remote=false), instead of a live cluster.
 	// Each file may hold multiple documents separated by "---" and/or a kind: List (or typed *List)
 	// wrapper whose items are expanded; a bare top-level JSON array is not supported. Duplicate
-	// resources are rejected. Not allowed together with Remote.
+	// resources are rejected. Not allowed together with Remote. Namespace scope is inferred from
+	// whether a resource sets metadata.namespace; namespaced resources must set it, otherwise
+	// namespace-scoped lookups ignore the requested namespace and match the resource regardless.
 	LocalLookupResourcesPaths []string
 	// NetworkParallelism limits the number of concurrent network-related operations (API calls, resource fetches).
 	// Defaults to DefaultNetworkParallelism if not set or <= 0.
@@ -155,10 +157,6 @@ func ChartLint(ctx context.Context, opts ChartLintOptions) error {
 
 	if opts.SecretKey != "" {
 		lo.Must0(os.Setenv("WERF_SECRET_KEY", opts.SecretKey))
-	}
-
-	if opts.Remote && len(opts.LocalLookupResourcesPaths) > 0 {
-		return fmt.Errorf("local lookup resources are not allowed together with remote mode")
 	}
 
 	if !opts.Remote {
--- a/pkg/action/chart_lint.go
+++ b/pkg/action/chart_lint.go
@@ -94,9 +94,11 @@ type ChartLintOptions struct {
 	// LocalKubeVersion specifies the Kubernetes version to use for linting when not connected to a cluster.
 	// Format: "major.minor.patch" (e.g., "1.28.0"). Defaults to DefaultLocalKubeVersion if not set.
 	LocalKubeVersion string
-	// LocalLookupResourcesPaths are paths to YAML/JSON manifest files whose resources the "lookup"
+	// LocalLookupResourcesPaths are paths to YAML or JSON manifest files whose resources the "lookup"
 	// template function resolves against in local mode (Remote=false), instead of a live cluster.
-	// Not allowed together with Remote.
+	// Each file may hold multiple documents separated by "---" and/or a kind: List (or typed *List)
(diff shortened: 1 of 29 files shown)


pkg/chart/lookup.go
```diff
@@ -0,0 +1,102 @@
...
+func NewLocalClientProvider(objects []*unstructured.Unstructured) *LocalClientProvider {
+	runtimeObjects := make([]runtime.Object, 0, len(objects))
+
+	registered := make(map[schema.GroupVersionKind]bool)
+
+	namespaced := make(map[schema.GroupVersionKind]bool)
+	for _, obj := range objects {
+		runtimeObjects = append(runtimeObjects, obj)
+		if obj.GetNamespace() != "" {
+			namespaced[obj.GroupVersionKind()] = true
```
ilya-lesikov, line 40: MAJOR: namespaced scope is inferred purely from whether a supplied stub object carries `metadata.namespace`, not from the kind's actual API scope. A namespaced kind stubbed without `metadata.namespace` is therefore treated as cluster-scoped, so a lookup like `lookup "v1" "Pod" "default" "pod1"` ignores the requested namespace and returns the object anyway — diverging from live-cluster behavior. Namespace omission in manifests is common (defaulted at apply time), so this can silently produce wrong template branching. The reverse error is also possible: a cluster-scoped kind stubbed with a stray namespace would be treated as namespaced. Consider rejecting namespace-less stubs for kinds used in namespaced lookups, or documenting that namespaced stub resources must set `metadata.namespace`.
(generated by pi-pi)

ilya-lesikov (commented):
Automated review of the `--lookup-resources` offline-lookup feature. 5 findings (2 major correctness gaps, 3 minor), all anchored to the diff below.

(generated by pi-pi)

pkg/chart/chart_render.go
```diff
@@ -272,6 +285,57 @@ func RenderChart(ctx context.Context, chartPath, releaseName, releaseNamespace s
...
+	for _, filePath := range paths {
+		content, err := os.ReadFile(filePath)
+		if err != nil {
+			return nil, fmt.Errorf("read file %q: %w", filePath, err)
+		}
+
+		for i, manifest := range releaseutil.SplitManifestsToSlice(string(content)) {
+			obj, _, err := scheme.Codecs.UniversalDecoder().Decode([]byte(manifest), nil, &unstructured.Unstructured{})
+			if err != nil {
+				return nil, fmt.Errorf("decode resource #%d for %q: %w", i+1, filePath, err)
```
ilya-lesikov, line 306: MINOR: the `resource #%d` number uses the post-split index, but the splitter drops empty/comment-only documents, so the reported number can differ from the document's literal position in the file. Additionally, top-level documents are reported 1-based (`#N`) while list-item errors are reported 0-based (`item N`) — inconsistent and potentially confusing when locating the offending resource.
(generated by pi-pi)

ilya-lesikov: MINOR: a bare top-level JSON array of resources is unsupported and fails here with an opaque Kubernetes decode error ("cannot unmarshal array into Go value ...") rather than a message explaining that top-level arrays are not accepted. A clearer error would help users who pass an array manifest.
(generated by pi-pi)

pkg/action/chart_lint.go
```diff
@@ -151,6 +157,10 @@ func ChartLint(ctx context.Context, opts ChartLintOptions) error {
 		lo.Must0(os.Setenv("WERF_SECRET_KEY", opts.SecretKey))
 	}
 
+	if opts.Remote && len(opts.LocalLookupResourcesPaths) > 0 {
+		return fmt.Errorf("local lookup resources are not allowed together with remote mode")
+	}
```
ilya-lesikov: hmm, let's do without this failure, just use opts.LocalLookupResourcesPaths only in local mode

pkg/action/chart_render.go
```diff
@@ -158,6 +164,10 @@ func ChartRender(ctx context.Context, opts ChartRenderOptions) (*ChartRenderResu
 		lo.Must0(os.Setenv("WERF_SECRET_KEY", opts.SecretKey))
 	}
 
+	if opts.Remote && len(opts.LocalLookupResourcesPaths) > 0 {
```
ilya-lesikov: same

pkg/lookup/lookup.go
```diff
@@ -0,0 +1,102 @@
+package lookup
```
ilya-lesikov: not a separate package for just a lookup

pkg/chart/lookup.go
dmmordvi: We are unable to determine whether the resource stub is namespaced or not in local mod. So documented that namespaced stubs must set `metadata.namespace`

ilya-lesikov (approved):

PR #664, opened 2026-07-21
feat: send untouched resource info to legacy progress reporter

dmmordvi:
<a href="https://cubic.dev/pr/werf/nelm/pull/664?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/action/common.go
+++ b/pkg/action/common.go
@@ -190,9 +190,10 @@ func runFailurePlan(ctx context.Context, releaseNamespace string, failedPlan *pl
 	log.Default.Debug(ctx, "Execute failure plan")
 
 	if err := plan.ExecutePlan(ctx, releaseNamespace, failurePlan, taskStore, logStore, informerFactory, history, clientFactory, plan.ExecutePlanOptions{
-		LegacyProgressReporter: opts.LegacyProgressReporter,
-		TrackingOptions:        opts.TrackingOptions,
-		NetworkParallelism:     opts.NetworkParallelism,
+		LegacyProgressReporter:   opts.LegacyProgressReporter,
+		TrackingOptions:          opts.TrackingOptions,
+		NetworkParallelism:       opts.NetworkParallelism,
+		InstallableResourceInfos: installableInfos,
 	}); err != nil {
 		critErrs.Add(fmt.Errorf("execute failure plan: %w", err))
 	}
--- a/pkg/legacy/progrep/progress_report.go
+++ b/pkg/legacy/progrep/progress_report.go
@@ -29,8 +29,9 @@ type ProgressReport struct {
 	StageReports []StageReport `json:"stageReports"`
 }
 
-// StageReport contains ALL operations in the plan -- from the very first report, every
-// operation is present (initially as Pending).
+// StageReport contains ALL operations in the plan, plus untouched resources reported as
+// NoOp with status Completed -- from the very first report, every operation is present
+// (initially as Pending).
 type StageReport struct {
 	Operations []Operation `json:"operations"`
 }
--- a/pkg/legacy/progrep/progress_report.go
+++ b/pkg/legacy/progrep/progress_report.go
@@ -13,6 +13,7 @@ const (
 	OperationTypeDelete         OperationType = "Delete"
 	OperationTypeApply          OperationType = "Apply"
 	OperationTypeRecreate       OperationType = "Recreate"
+	OperationTypeNoOp           OperationType = "NoOp"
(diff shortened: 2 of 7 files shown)


pkg/action/common.go
```diff
@@ -190,9 +190,10 @@ func runFailurePlan(ctx context.Context, releaseNamespace string, failedPlan *pl
 	log.Default.Debug(ctx, "Execute failure plan")
 
 	if err := plan.ExecutePlan(ctx, releaseNamespace, failurePlan, taskStore, logStore, informerFactory, history, clientFactory, plan.ExecutePlanOptions{
-		LegacyProgressReporter: opts.LegacyProgressReporter,
-		TrackingOptions:        opts.TrackingOptions,
-		NetworkParallelism:     opts.NetworkParallelism,
+		LegacyProgressReporter:   opts.LegacyProgressReporter,
+		TrackingOptions:          opts.TrackingOptions,
+		NetworkParallelism:       opts.NetworkParallelism,
+		InstallableResourceInfos: installableInfos,
```
dmmordvi, line 196: The full original installableInfos is passed into the failure plan, so every install resource still present in the cluster is re-emitted as NoOp/Completed in a stage where it is not an operation of that plan. This diverges from the StageReport doc contract ("ALL operations in the plan"). If this is intended (giving the consumer the full picture), consider clarifying the StageReport doc comment to mention these NoOp pseudo-operations.
(generated by pi-pi)

pkg/legacy/progrep/progress_report.go
```diff
@@ -13,6 +13,7 @@ const (
 	OperationTypeDelete         OperationType = "Delete"
 	OperationTypeApply          OperationType = "Apply"
 	OperationTypeRecreate       OperationType = "Recreate"
+	OperationTypeNoOp           OperationType = "NoOp"
```
dmmordvi, line 16: OperationTypeNoOp is a new enum value that is serialized over the wire to an external consumer. If the consumer validates the operation type against a fixed set or uses an exhaustive switch, an unknown "NoOp" value could break parsing or render incorrectly. Consider adding a test/note in this repo that pins the set of valid wire values so compatibility is guaranteed rather than relying on the consumer tolerating unknown types.
(generated by pi-pi)

dmmordvi (commented):
Code review (3 findings, all MINOR — no blockers). The PR is correct and ready to merge; the two potentially serious points below are accepted as deliberate.

Findings not anchorable to the diff:
pkg/plan/legacy_progress_reporter_ai_test.go:— No test covers the real main→failure/rollback scenario where the full instResInfos is re-passed as untouched into a second StartStage; only the nil-untouched case is covered by TestAI_StartStage_UntouchedScopedToStageAndFrozen. Recommend adding a scenario test that pins the expected re-emission behavior so regressions become visible. (generated by pi-pi)

pkg/plan/legacy_progress_reporter.go
```diff
@@ -90,6 +104,33 @@ func (r *LegacyProgressReporter) startStage(p *Plan, resolvedNamespaces map[stri
 			})
 		}
 
+		for _, info := range untouched {
```
dmmordvi, line 107: A resource with real drift can reach this NoOp/Completed branch. resourceInstallType() returns ResourceInstallTypeNone not only when the live and dry-apply states match, but also when a resource policy (skip-update / skip-create / skip-recreate) forces None despite an actual diff (skippedByPolicy=true). Such a resource produces no plan operation (and mustTrackReadiness is false), so it flows into this loop and is reported as NoOp/Completed. That reads as "unchanged" to the user even though the resource actually differs from the desired state and was only skipped by policy. This is display-only in the legacy progress reporter (deploy behavior is unaffected), but the label is misleading; consider distinguishing "skipped by policy" from "unchanged".
(generated by pi-pi)

dmmordvi (commented):
Code review pass (pi-pi). One new MINOR finding below. A prior review already flagged the OperationTypeNoOp wire-compatibility concern on pkg/legacy/progrep/progress_report.go:16, so it is not re-posted here.

pkg/plan/legacy_progress_reporter.go
dmmordvi: We return NoOp type, so it is enough

PR #675, opened 2026-07-30
feat: embed deno binary into release binaries behind embeddeno tag

dmmordvi:
## Summary

Release binaries ship the pinned deno 2.9.4 inside them instead of downloading it from GitHub on first use of a TypeScript chart. Embedding is gated behind the `embeddeno` build tag, so only `build:dist:*` targets carry the blob; untagged dev builds keep the runtime-download behavior and stay small. What makes the shipped binary the deno this repository chose is a lock file committed here, not a checksum fetched alongside the download.

## Key changes

- **`pkg/ts/denolock/data/lock.json`** — committed and embedded: the deno version and, per platform, the target triple plus the sha256 of the release archive and of the binary inside it. The only place the deno version is configured, replacing the `denoVersion` constant. The download path verifies the archive and then the unpacked binary against it, so `fetchExpectedChecksum` is gone and nothing takes a digest from the network any more.
- **`scripts/denolock`** generates the lock — downloads every archive, hashes it, unpacks and hashes the binary, cross-checks each digest against the upstream `.sha256sum` before writing. Wired as `task deno:lock` (regenerate, or bump with `-- -version 2.8.0`), `task lint:deno-lock` (offline, part of `task lint`) and `task check:deno-lock-upstream` (network, by hand).
- **Runtime extraction** — `extractEmbeddedDeno` streams gzip into a content-addressed cache dir while hashing, rejects a mismatch, then chmods and atomically renames, under a flock with a post-lock double-check. Never holds the ~90 MB decompressed binary in memory. Carries no build tag, because werf imports it regardless of its own tags.
- **Per-platform embedding** — five `pkg/ts/embed_<goos>_<goarch>.go` files rely on Go's implicit filename constraints, so each binary embeds only its own ~42-47 MiB blob.
- **Embedded bytes over context** — `TypeScriptOptions` gained `EmbeddedDenoCompressed`, letting werf ship its own deno without a build tag (build tags do not reach a dependency whose blob lives in the read-only module cache). No digest travels with it: those bytes can only be the release the lock pins, so nelm looks the digest up itself. Resolution order in `getDenoBinary`: explicit `binaryPath` → ctx-supplied bytes → own `embeddeno` blob → download, extraction lazy behind `hasTSFiles`.
- **`scripts/embed-deno`** downloads the pinned release for one platform through the verifying path and gzips it into `pkg/ts/embed/<os>/<arch>/deno.gz`. Driven by `task deno:embed`, and by werf as `go run github.com/werf/nelm/scripts/embed-deno`. `DownloadDenoForPlatform` is exported for it — the only export this branch adds to `pkg/ts`.
- **Fixes a pre-existing windows bug** — the checksum parser took the first space-separated field, but deno publishes the windows checksum as PowerShell `Get-FileHash` output, so it extracted the literal `"Algorithm"` and errored. The digest is located by shape instead. The parser now lives in `scripts/denolock`, the only code left that reads an upstream checksum.
- **Taskfile** — `deno:embed` generates only the platform being built, fingerprinted per platform via `label`. Also fixes `_build:dev`/`_build:dist`, which passed `-tags="{{.ctags}}"`, a variable defined nowhere, making `-tags` permanently empty. wormatter is excluded from `pkg/ts/embed_*.go`: it drops the `//go:embed` directive of a single `var`, which compiles fine and silently yields an empty embedded deno.

## Why

A self-contained binary removes a network dependency from the first TypeScript render: today the first `chart ts build` on a fresh machine pulls ~45 MiB from GitHub, which fails in air-gapped environments and in CI without egress to `github.com`.

Pinning the digests is what makes that binary trustworthy. Verifying a download against `<archive>.sha256sum` catches a truncated transfer and nothing else — it comes from the same host, over the same connection, moments before the artifact it vouches for. Hashing the download on the release machine and embedding that digest next to the blob is circular for the same reason. Committing the digests moves the trust anchor into the repository the way `go.sum` does: what a binary verifies against was reviewed in a diff and is in git history.

The windows checksum fix is a prerequisite — the lock cannot be generated for windows without it, and it means the runtime download was broken on windows/amd64 independently of embedding.

## Verification

- Generated the lock from scratch: every archive digest agreed with the `.sha256sum` deno publishes for it, and each binary digest was reproduced independently by hashing the blob the embed step wrote.
- `task check:deno-lock-upstream` — all five assets still match the lock. It also reported 2.9.4 as available while 2.7.1 was pinned, which is how the pin got bumped: `task deno:lock -- -version 2.9.4` rewrote the lock, and the drift check has been clean on it since.
- Tampered with one `archiveSha256` and re-ran `task deno:embed`: failed on the mismatch, no blob written.
- Confirmed a tagged build's embedded blob is populated (43,953,681 bytes, darwin/arm64): the wormatter behavior the exclusion works around produces no compile error, only an empty blob.
- `task build:dist:all version=0.0.0` from a clean tree: five platforms embedded once each, five binaries produced. Size delta with identical ldflags: tagged − untagged = 42.2 MiB against a 41.9 MiB blob.
- Offline end-to-end, cold cache, proxies pointed at a dead port: a tagged darwin/arm64 binary bundled a TypeScript chart; an untagged build under the same block failed with a download error, so the fallback path is still live. `--deno-binary-path` created no cache dir at all.
- macOS exec trust: the extracted deno is byte-identical to the official download (`cmp` clean), `codesign -v` valid with hardened runtime intact. `spctl` returns the same `Unnotarized Developer ID` verdict for the pristine download, so extraction introduces no new rejection.
- werf built against this branch with its companion change and `GOWORK=off`, i.e. against the published module rather than the workspace copy: its blob hashes to the binary digest the lock pins.
- **Not run:** `task verify:binaries:dist:all` (`scripts/verify-dist-binaries.sh` needs bash ≥ 4, macOS ships 3.2 — applied its regexes by hand instead); first launch on a clean macOS without developer tooling; any Windows Defender/SmartScreen check.

## Review focus / risks

- **The lock is only as strong as the review of its diff.** A bump changes ten digests at once; waving them through makes the mechanism decorative. That is the one process assumption this design makes.
- **A release already malicious when pinned is not covered.** Only upstream provenance would help, and deno publishes none — the release assets are archives and `.sha256sum` files, nothing signed.
- **No scheduled check for a replaced release asset.** GitHub allows replacing one without moving the tag, but a release build already fails closed on the checksum, so a cron would only buy earlier notice of a rare event. `task check:deno-lock-upstream` exists to diagnose that failure — replaced asset vs broken download, in five requests.
- **werf#7725 must land with this.** werf builds against a pseudo-version of this branch and consumes `scripts/embed-deno` and `TypeScriptOptions`, so it has to follow every shape they take. Its companion PR is pinned to the head of this branch and takes the expected digest from `denolock`, which also puts werf's blob under the same reviewed pin instead of one it computed from its own download.
- **Release builds still need egress to `github.com` at build time** and every binary grows by ~42-47 MiB (gzip is the floor — deno's zips are already deflate-compressed). The `archiveSha256` is a ready cache key, but no caching is wired up here.
- Extraction trusts an existing cache file on `os.Stat` alone. Sound because the directory name is the content hash and files only land there via atomic rename, but worth a second opinion.
- The wormatter exclusion works around a formatter bug with a silent failure mode; it should go when that is fixed.
- AI-written tests need `task test:unit tags=ai_tests` — plain `task test:unit` does not compile them.

<a href="https://cubic.dev/pr/werf/nelm/pull/675?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pkg/ts/downloader.go
```diff
@@ -24,7 +24,26 @@ import (
 
 const denoVersion = "2.7.1"
```
dmmordvi: MAJOR — The embedded Deno is pinned by version only, not by a fixed content digest. At build time the generator downloads both the release archive and its checksum from the same upstream (`github.com/denoland/deno`), and the generated blobs are gitignored and regenerated per build. The checksum therefore only protects against transit corruption, not against a compromised or mutated upstream release — a matching malicious archive+checksum would be embedded silently. Consider committing a known-good per-platform sha256 into the repo and verifying the generator output against it, so the version pin becomes a real artifact pin.

(generated by pi-pi)

pkg/ts/embedded_deno.go
```diff
@@ -0,0 +1,115 @@
...
+		return "", fmt.Errorf("unexpected embedded deno checksum format: %q", expectedSHA256)
+	}
+
+	cacheDir := helmpath.CachePath("nelm", "deno", "embedded-"+expectedSHA256[:16])
+	if err := os.MkdirAll(cacheDir, 0o755); err != nil {
+		return "", fmt.Errorf("create cache directory for embedded deno: %w", err)
+	}
+
+	denoPath := filepath.Join(cacheDir, denoBinaryName(runtime.GOOS))
+	if _, err := os.Stat(denoPath); err == nil {
```
dmmordvi, line 37: MINOR — A cache hit is trusted on `os.Stat` alone: it accepts any existing entry without checking that it is a regular file (not a symlink) and without re-verifying its content. The content-addressed directory name, atomic rename and flock make the normal concurrent extraction path sound, but they do not protect against a file or symlink pre-planted at this predictable, user-writable path (local cache poisoning; relevant with shared or misconfigured cache homes). The directory key also uses only the first 16 hex chars of the digest. Cheap hardening: reject non-regular cache files before returning them.

(generated by pi-pi)

pkg/ts/embedded_deno.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return fmt.Errorf("create temp file in %s: %w", destDir, err)
+	}
+
+	tmpPath := tmpFile.Name()
+	defer os.Remove(tmpPath)
+
+	hasher := sha256.New()
+
+	if _, err := io.Copy(io.MultiWriter(tmpFile, hasher), gzReader); err != nil {
```
dmmordvi, line 91: MINOR — `extractDeno` copies the decompressed gzip stream via `io.Copy` with no size limit. The sha256 check does not bound this, since the caller supplies both the compressed bytes and the expected digest. For the trusted embedded blob this is fine, but `ExtractEmbeddedDeno` and the `EmbeddedDenoCompressed` option are exported API: if that input ever comes from a less-trusted source, a crafted gzip could exhaust disk before the hash is even evaluated. Consider bounding extraction with an `io.LimitReader` sized to the expected Deno binary.

(generated by pi-pi)

pkg/ts/downloader.go
```diff
@@ -119,14 +138,38 @@ func fetchExpectedChecksum(ctx context.Context, httpClient *resty.Client, archiv
...
 
+func denoBinaryName(goos string) string {
+	return lo.Ternary(goos == "windows", "deno.exe", "deno")
+}
+
+// findChecksum extracts a sha256 hex digest from a checksum file. Deno
+// publishes plain "<hex>  <file>" for unix targets, but PowerShell
+// Get-FileHash output ("Hash : <UPPERCASE HEX>") for windows ones, so the
+// digest is located by shape rather than by field position.
+func findChecksum(body string) (string, bool) {
```
dmmordvi: MINOR (nit) — `findChecksum` returns the first matching 64-char hex token in the body, not a uniquely identified one. This is correct for the actual Deno checksum files (unix `sha256sum` puts the hex first; the Windows PowerShell `Get-FileHash` output lists `Hash :` before `Path :`), so there is no functional bug. Flagging only because the accompanying wording calls it “the single 64-char hex token” — “first” would be accurate. No code change required.

(generated by pi-pi)

dmmordvi (commented):
Automated code review of the embedded-Deno change: 5 findings — 1 MAJOR (build-time artifact pinning) and 4 MINOR (2 nits). All are non-blocking; the runtime trust model does not regress relative to the pre-existing download path. Four findings are attached as line comments below.

Findings not anchorable to the diff:

pkg/ts/deno.go:— — MINOR (nit): `RunDenoInstall` calls `getDenoBinary` unconditionally, without the `hasTSFiles` gate used by the bundle and render paths. This is not a user-visible bug — its caller initializes TypeScript boilerplate, which needs Deno — but Deno extraction is therefore not universally gated behind a TypeScript-presence check as one might expect from the surrounding code.

(generated by pi-pi)

pkg/ts/embedded_deno.go
dmmordvi: i think, we can ignore it

pkg/ts/embedded_deno.go
dmmordvi: i think, we can ignore it

ilya-lesikov (approved):

PR #686, opened 2026-08-07
fix(helm): serialize action configuration initialization

alexey-igrychev:
## Summary

Concurrent Helm command initialization can mutate one shared `action.Configuration` at the same time. `Configuration.Init` now serializes these updates on the Nelm `2` branch used by werf `3`.

## What

- `Configuration.Init` protects client, storage, and hook-output replacement with its existing configuration-local lock.
- AI regression test starts two concurrent memory-driver initializations of one configuration.
- VERIFIED: `task --yes test:ginkgo paths="./pkg/helm/pkg/action" tags="ai_tests" parallel=false -- --race` passes.

## Why

Race-instrumented werf e2e compose testing reported concurrent writes at `Configuration.Init` while Helm commands initialized the same configuration. This caused werf to exit with code 66.

Tracked in werf/werf#7775.

### DIFF
--- a/pkg/helm/pkg/action/action.go
+++ b/pkg/helm/pkg/action/action.go
@@ -663,6 +663,9 @@ func (cfg *Configuration) recordRelease(r *release.Release) {
 
 // Init initializes the action configuration
 func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namespace, helmDriver string) error {
+	cfg.mutex.Lock()
+	defer cfg.mutex.Unlock()
+
 	kc := kube.New(getter)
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- `task --yes build`.
- `task --yes test:ginkgo paths="./pkg/helm/pkg/action" tags="ai_tests" parallel=false -- --race`.
- Mutation: removed `Configuration.Init` locking; race test reported concurrent initialization. Restored locking; test passed.

## Review focus

- Confirm the existing `Configuration` lock serializes every field update in `Init` without changing branch `2` initialization semantics.

Issue #677, opened 2026-08-01
Secrets: preserve YAML scalar types and switch to authenticated encryption without breaking existing data


alexey-igrychev:
### Problem

werf's secret encryption has three defects that are all rooted in `werf/common-go/pkg/secret` (references below are against `v0.0.0-20260504183956-43da716392f7`).

**1. YAML scalar types are lost on encryption**

`yaml_encoder.go:214` encrypts every scalar as `fmt.Sprintf("%v", value)`, so the YAML tag is never persisted. Decryption always restores `!!str`. Encrypting `foo: 123` and decrypting it yields `foo: "123"`. The code carries two explicit `FIXME`s at `:200` and `:213`, the latter stating that this is a compatibility mode with a previous werf version.

Reported in #71 and #80.

**2. Block scalar style and comments are lost**

`deepCopyNode` preserves `Style` and the comment fields, but `node.Encode(...)` at `:192` and `:219` overwrites the node wholesale and discards them. A `>-` block scalar comes back as a plain string.

Reported in #74.

**3. Encryption is unauthenticated**

`aes_encoder.go` uses AES-CBC with a random IV and no MAC, so there is no integrity check: a corrupted or tampered ciphertext is not reliably detected, and a wrong key produces garbage rather than a clear error. Related weaknesses: `unpad` (`:113-122`) does not validate padding contents and does not reject `0` or values `> BlockSize`, and `IsExtractDataError` (`:133-146`) distinguishes error classes by matching message prefixes.

Note the key is 16 random bytes (`GenerateAesSecretKey`, `:19-28`), i.e. AES-128.

### Solution (if you have one)

All three can be fixed **without breaking existing encrypted data**, because the ciphertext format already contains an unused version field.

`Encrypt` (`:57-62`) writes a 2-byte little-endian prefix holding `aes.BlockSize`, and `Decrypt` (`:80-89`) reads past it but ignores the stored value — the IV size is hardcoded. That field can be repurposed as a format version: `16` means the existing CBC format, and a reserved high range (`0xF000`-`0xFFFF`) is used for new versions. Old ciphertext keeps decrypting exactly as today; unknown versions produce a clear error.

**Work items**

1. **Format version discriminator** — read the prefix, add named constants, document the format contract in the package `doc.go`.
[...]

PR #706, opened 2026-08-31
feat: support authenticated secret values

alexey-igrychev:
## Summary

Nelm v2 now uses common-go branch `1`, adding authenticated AES-GCM secret encryption while retaining read support for existing AES-CBC secrets. It can read and write the format required by the werf v3 rollout.

Related: https://github.com/werf/nelm/issues/677

### Breaking

- BREAKING: secrets written by nelm v2 use AES-GCM formats that cannot be read by earlier nelm or werf releases. Upgrade every CI job and saved deploy-plan consumer before re-encrypting a repository.
- Existing AES-CBC encrypted values remain readable, but new writes use authenticated formats.

### Secret values

- Whole-blob secrets use version 2 and YAML scalar secrets use version 3, preventing whole-blob values from being interpreted as scalar metadata.
- YAML scalar tags, block styles, and value comments survive a secret-values edit cycle. Existing values that lost their original type remain strings until re-entered.
- Whole-blob ciphertext can be used as a scalar in an additional secret-values file without frame corruption.

## Why

The previous common-go AES-CBC format accepted tampered data and wrong keys without reliable detection, while secret YAML encryption discarded scalar metadata. Nelm needs the common-go v1 implementation so its v2 release participates in the coordinated authenticated-secret rollout with werf v3.

<a href="https://cubic.dev/pr/werf/nelm/pull/706?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,7 +57,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	github.com/tidwall/sjson v1.2.5
 	github.com/wI2L/jsondiff v0.7.0
-	github.com/werf/common-go v0.0.0-20260428201303-b0dcadceca5c
+	github.com/werf/common-go v0.0.0-20260831162604-46c3b45163be
 	github.com/werf/kubedog v0.13.1-0.20260807153813-e8f61a4bc90a
 	github.com/werf/lockgate v0.1.1
 	github.com/werf/logboek v0.6.1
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- Linux (Go 1.25.5): `task build` and `task test:unit` passed against this commit.
- `task format` and `task lint` both fail with the same 12 pre-existing gosec G602 findings on `pkg/action/release_history.go` and `pkg/action/release_list.go` when run from origin/2; the bump adds no diff beyond go.mod/go.sum.

## Review focus

- The common-go version range includes the versioned AES-GCM secret-format rollout and the removal of an unused deprecated missed-secret-key mode. Nelm v2 compiles and tests without that removed API.

## Follow-up

- [ ] BLOCKER: Update github.com/werf/common-go in werf v3 to the same common-go/1 commit, then verify persisted deploy-plan and secret workflows.
- [ ] Document the minimum nelm v2 and werf v3 releases needed to read AES-GCM secret values before release.
- [ ] Close https://github.com/werf/nelm/issues/677 manually after both downstream bumps land.

